Issue #272, opened 2022-01-21
An API endpoint is needed for fetching pages of competencies


davidvandusen:
## Expected behaviour

An HTTP `GET` request to `/competencies` responds with a paginated list of competencies from the database.

## Actual behaviour

There is no way to fetch competencies from the database via the API.

## Details and resources

There are helpers on the backend (middleware) for handling paginated collections.

There is a pattern of having database queries in the services layers, and validation in the router layer.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #285, opened 2022-01-27
Add competencies services and test 

BRashad:
## Proposed changes

This change creates a competencies service function and tests to fetch competencies list with pagination. 

This is the 1st PR covering the query portion. for ticket #272 

## Checklist

### DIFF
--- a/api/src/services/__tests__/competenciesService.ts
+++ b/api/src/services/__tests__/competenciesService.ts
@@ -2,18 +2,21 @@ import { listCompetencies } from '../competenciesService';
 import { mockQuery } from '../mockDb';
 
 describe('listCompetencies', () => {
-  it('should load a page that lists the competencies in the database with their titles and descriptions.', async () => {
+  it('should query the lists of the competencies in the database with their titles and descriptions.', async () => {
+    const competencies = [
+      {
+        id: 2,
+        label: 'label',
+        description: 'description',
+      },
+    ];
+    const limit = 3;
+    const offset = 4;
     mockQuery(
       'select `id`, `label`, `description` from `competencies` limit ? offset ?',
-      [10, 10],
-      [{ id: 1, label: 'Code', description: 'To code or not to code...' }]
+      [limit, offset],
+      competencies
     );
-    expect(await listCompetencies(10, 10)).toEqual([
-      {
-        id: 1,
-        label: 'Code',
-        description: 'To code or not to code...',
-      },
-    ]);
+    expect(await listCompetencies(limit, offset)).toEqual(competencies);
   });
 });
--- a/api/src/services/__tests__/competenciesService.ts
+++ b/api/src/services/__tests__/competenciesService.ts
@@ -1,0 +1,19 @@
+import { listCompetencies } from '../competenciesService';
+import { mockQuery } from '../mockDb';
+
+describe('listCompetencies', () => {
+  it('should load a page that lists the competencies in the database with their titles and descriptions.', async () => {
+    mockQuery(
+      'select `id`, `label`, `description` from `competencies` limit ? offset ?',
+      [10, 10],
+      [{ id: 1, label: 'Code', description: 'To code or not to code...' }]
+    );
+    expect(await listCompetencies(10, 10)).toEqual([
+      {
+        id: 1,
+        label: 'Code',
+        description: 'To code or not to code...',
+      },
+    ]);
+  });
+});
--- a/api/src/services/__tests__/competenciesService.ts
+++ b/api/src/services/__tests__/competenciesService.ts
@@ -1,22 +1,24 @@
 import { listCompetencies } from '../competenciesService';
 import { mockQuery } from '../mockDb';
 
-describe('listCompetencies', () => {
-  it('should query the lists of the competencies in the database with their titles and descriptions.', async () => {
-    const competencies = [
-      {
-        id: 2,
-        label: 'label',
-        description: 'description',
-      },
-    ];
-    const limit = 3;
-    const offset = 4;
-    mockQuery(
-      'select `id`, `label`, `description` from `competencies` limit ? offset ?',
-      [limit, offset],
-      competencies
-    );
-    expect(await listCompetencies(limit, offset)).toEqual(competencies);
+describe('competenciesService', () => {
+  describe('listCompetencies', () => {
+    it('should query the lists of the competencies in the database with their labels and descriptions', async () => {
+      const competencies = [
+        {
+          id: 2,
+          label: 'label',
+          description: 'description',
+        },
+      ];
+      const limit = 3;
+      const offset = 4;
+      mockQuery(
+        'select `id`, `label`, `description` from `competencies` limit ? offset ?',
+        [limit, offset],
+        competencies
+      );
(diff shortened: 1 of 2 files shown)


api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,19 @@
+import { listCompetencies } from '../competenciesService';
+import { mockQuery } from '../mockDb';
+
+describe('listCompetencies', () => {
+  it('should load a page that lists the competencies in the database with their titles and descriptions.', async () => {
```
pthan1, line 5: Should this test description be re-written to describe what is actually being tested?

api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,19 @@
...
+    mockQuery(
+      'select `id`, `label`, `description` from `competencies` limit ? offset ?',
+      [10, 10],
```
KanishDabreo, line 8: What was your reason for using 10,10  for the limit and offset?

api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,19 @@
...
+      [{ id: 1, label: 'Code', description: 'To code or not to code...' }]
```
KanishDabreo, line 9: Why is the label value set to 'Code'?

api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,19 @@
...
+    );
+    expect(await listCompetencies(10, 10)).toEqual([
+      {
+        id: 1,
+        label: 'Code',
+        description: 'To code or not to code...',
```
KanishDabreo, line 15: Why is the description value set to 'To code or not to code...'?

pthan1, line 15: Would it be valuable to test a real-world example of a comptency here instead of "Code"?

api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,19 @@
+import { listCompetencies } from '../competenciesService';
+import { mockQuery } from '../mockDb';
+
+describe('listCompetencies', () => {
```
pthan1, line 4: Would it be better to change "listCompetencies" to "competenciesService"?  This test file should test other service functions in the future.

api/src/services/__tests__/competenciesService.ts
BRashad: 1st 10 refers to the "code injection"/? after "limit" to limit the number of competencies requester at a time
2nd 10 refers to "code injection"/? after "offset" to consider competencies listed previously.

api/src/services/__tests__/competenciesService.ts
BRashad: "Label" is the title/name of the competency.

api/src/services/competenciesService.ts
```diff
@@ -0,0 +1,10 @@
+import db from './db';
+
+export const listCompetencies = async (limit: number, offset: number) => {
+  const competencies = await db('competencies')
+    .select('id', 'label', 'description')
```
pthan1, line 5: The competencies tables has other columns.  Should the function also retrieve those columns?

api/src/services/__tests__/competenciesService.ts
BRashad: This is a fake description of the fake competency. Kind of placeholder text

EnriqueTMT: Might be a Shakespeare reference. lol

BRashad: We shall change it to general statements like: 
label: "Competency name goes here"
description: "Competency description goes here"

api/src/services/__tests__/competenciesService.ts
BRashad: it was initially competenciesService. David suggested following patterns in other files in naming the functions.

api/src/services/competenciesService.ts
BRashad: Yes. Please let me know what are those columns.

api/src/services/__tests__/competenciesService.ts
BRashad: yes, agree. New comment: should query for the lists of competencies in the database with their titles and descriptions

api/src/services/__tests__/competenciesService.ts
pthan1: I'm looking at other files and I think the convention set is to have the values without the "goes here" text.

davidvandusen: Why not follow the pattern from other files? https://github.com/OpenTree-Education/rhizone-lms/blob/main/api/src/services/__tests__/reflectionsService.ts#L38

api/src/services/__tests__/competenciesService.ts
davidvandusen: The patterns in the other files are that there it a top level `describe` block with the name of the module, and then an inner `describe` block for each export from that module.

api/src/services/competenciesService.ts
pthan1: created_at

api/src/services/__tests__/competenciesService.ts
```diff
@@ -0,0 +1,22 @@
...
+  it('should query the lists of the competencies in the database with their titles and descriptions.', async () => {
```
davidvandusen, line 5: This test description states that "titles" are queried for, but the test doesn't have anything in it called "title". Is this supposed to be "labels"?

Also, test descriptions typically do not have a period at the end.

api/src/services/__tests__/competenciesService.ts
BRashad: Yes we now follow the pattern

BRashad: Yes we now follow the pattern

BRashad: Yes we now follow the pattern

BRashad: Yes we now follow the pattern

api/src/services/__tests__/competenciesService.ts
BRashad: Just made changes as suggested.  We added a top-level 'describe' block.

api/src/services/__tests__/competenciesService.ts
BRashad: Yes, it was supposed to be "labels". We changed it now to 'labels'. Also removed full stop.

api/src/services/competenciesService.ts
BRashad: We deleted created_at.

KanishDabreo (approved):

api/src/services/competenciesService.ts
pthan1: I agree. I don't think we need to retrieve those columns right now.  I noticed that the ReflectionsService selected for those columns.

pthan1 (approved):
LGTM!

Issue #292, opened 2022-02-07
rhi.zone/competencies not viewable from production but is in dev


EnriqueTMT:
## Expected behaviour

Navigating to `rhi.zone/competencies` loads a page that lists the competencies in the database with their titles and descriptions.

## Actual behaviour

Navigating to `rhi.zone/competencies` loads "Page Not Found" via the address bar or refresh.

## Details and resources

Fixing a bug from issue #273 and pull request #287.
Navigating to Terms of Use and Privacy Policy in `rhi.zone-development` loads page whereas in `rhi.zone', it does not. It says "Page Not Found" (see attached screenshot).
Navigating to `rhi.zone-development/competencies`, when there are no competencies, it says "There are no competencies."

![image](https://user-images.githubusercontent.com/75624926/152878949-2b929b20-5e8c-45de-b09f-eae11a522077.png)

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

Issue #90, opened 2021-11-09
Deployment requires more than a single button press


davidvandusen:
## Expected behaviour

Deploying changes to the services requires a single button press.

## Actual behaviour

Deploying each service requires a separate list of manual steps.

## Details and resources

Eventually we want to get to CI with Docker. Until then, we need a faster way to ship. This should be as simple as writing a script that can be run on the production EC2 box.

Depends on #51.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #293, opened 2022-02-08
Build and push a production webapp image to replace Netlify webapp

davidvandusen:
## Proposed changes

Serve the webapp from its own container in production. Uses `try_files`, falling back to `index.html`, to address #292.

After this is shipped, `webapp/netlify.toml` can be removed and that site deleted, since the app will be built and served from our infrastructure instead.

Another step along the way to addressing #90.

## Checklist

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -20,6 +20,7 @@ services:
   nginx:
     depends_on:
       - api
+      - webapp
     image: ghcr.io/opentree-education/rhizone-lms_nginx:latest
     ports:
       - "80:80"
@@ -28,3 +29,5 @@ services:
       - /etc/letsencrypt/:/etc/letsencrypt/
   redis:
     image: redis:6.2-alpine
+  webapp:
+    image: ghcr.io/opentree-education/rhizone-lms_webapp:latest
--- a/nginx/production/nginx.conf
+++ b/nginx/production/nginx.conf
@@ -28,6 +28,10 @@ http {
     # for more information.
     include /etc/nginx/conf.d/*.conf;
 
+    upstream rhizone_lms_webapp {
+        server webapp:80;
+    }
+
     server {
         listen       80;
         listen       [::]:80;
@@ -47,9 +51,7 @@ http {
         } # managed by Certbot
 
         location / {
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header Host rhizone.netlify.app;
-            proxy_pass https://rhizone.netlify.app;
+            proxy_pass http://rhizone_lms_webapp;
         }
     }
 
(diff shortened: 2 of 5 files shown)


webapp/default.conf
```diff
@@ -0,0 +1,10 @@
+server {
+    listen       80 default_server;
+    server_name  localhost;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html;
+        try_files $uri /index.html;
```
davidvandusen, line 8: This addresses #292 by returning `index.html` if the requested path isn't an existing file.

nginx/production/nginx.conf
```diff
@@ -47,9 +51,7 @@ http {
         } # managed by Certbot
 
         location / {
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header Host rhizone.netlify.app;
-            proxy_pass https://rhizone.netlify.app;
+            proxy_pass http://rhizone_lms_webapp;
```
davidvandusen, line 54: This is where we point to the new webapp container instead of Netlify.

docker-compose.production.yml
```diff
@@ -28,3 +29,5 @@ services:
       - /etc/letsencrypt/:/etc/letsencrypt/
   redis:
     image: redis:6.2-alpine
+  webapp:
```
davidvandusen, line 32: Start a new webapp service in the stack to point nginx to.

eileenxue (approved):
Everything looks good to me. 

What's the difference with serving the webapp on Nginx or Netflify?

webapp/default.conf
daveytech: To clarify: This copies the default config from webapp and sets nginx config through the docker build? Previously, was nginx config set without the try_files line?

davidvandusen: Previously the `nginx` service was forwarding requests for the webapp in production to Netlify using `proxy_pass`.

Now, instead of building on Netlify, the webapp gets built and deployed in an nginx container on the swarm that everything else runs in.

It would have been possible to set up redirect rules in Netlify to behave the same as `try_files` here, but I'm trying to reduce the number of technologies maintainers need to know to work on the infrastructure.

Does that clarify things?

daveytech: Yes it does. Thank you so much and the rest of the PR was well explained.

daveytech (approved):
Thank you for the well explained comments with the file changes.

davidvandusen:
@eileenxue There's very little difference. In both cases, the nginx service just forwards requests to another web server. There are advantages and disadvantages of each approach, and neither are perfectly ideal, but I think serving the webapp from a container in the stack is conceptually easier to maintain, matches better with development, and reduces the number of moving parts in the deploy process.

PR #296, opened 2022-02-08
Remove netlify config since webapp is published in an image now

davidvandusen:
## Proposed changes

In simplifying the deployment, the webapp is published in an image that is run on the Docker stack now, so there's no need for any of the Netlify config. The Netlify site that previously hosted the webapp has been deleted.

Just another step in #90.

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,6 +8,5 @@
 /docker-compose.production.yml @OpenTree-Education/production-engineers
 /webapp/.dockerignore @OpenTree-Education/production-engineers
 /webapp/Dockerfile @OpenTree-Education/production-engineers
-/webapp/netlify.toml @OpenTree-Education/production-engineers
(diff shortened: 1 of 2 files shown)


pthan1 (approved):
What'll be the next step to address issue #90 now that netlify is no longer being used?

davidvandusen:
@pthan1 The next step is to containerize the Flyway migrations. Currently they're done manually on prod.

eileenxue (approved):

eileenxue:
If you were to create another web app now, would you set up everything in Docker from the beginning?

davidvandusen:
@eileenxue It was always the case that we wanted to have deployments be fully containerized. The reason it's not set up that way now is because of externalities like timelines and budgets and other unknown unknowns.

The way priorities panned out, it was better to do the simplest thing to ship the software (that is, take on tech debt) and then invest time in doing things a better way when time is available (that is, repay the debt).

Issue #275, opened 2022-01-21
Competencies cannot be added to the database from the webapp


davidvandusen:
## Expected behaviour

A form in the webapp creates a new competency and links it to the current user's competency model.

## Actual behaviour

Competencies need to be added directly to the database or by hitting the API directly.

## Details and resources

The form could appear on the `/competencies` page. It should have separate text fields for entering the title and description of the competency.

The user experience could be improved by clearing out the form after it submits successfully to allow for another competency to be easily added.

The list of all competencies should be updated after successfully inserting the new one.

Success and error messages are presented depending on the outcome of the request.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

BRashad:
I have only one question: Have you tested simulated the server-side error in case the user competency input can not be successfully inserted into DB ?

BRashad:
on my local machine, all work as per the ticket requirements. Subject to question feedback on server-side error upon unsuccessful competency insert into DB. the case is good to be merged. 

![Screen Shot 2022-02-08 at 2 29 12 PM](https://user-images.githubusercontent.com/10593476/153078854-a715a6ea-05fd-4d15-b07b-73b12c0ff98f.png)
![Screen Shot 2022-02-08 at 2 29 26 PM](https://user-images.githubusercontent.com/10593476/153078860-ff931589-355f-49be-88e3-1c6d33884036.png)

PR #297, opened 2022-02-08
Add competency form so that users can add new competencies to the database

eileenxue:
## Proposed changes

- Make `CreateCompetencyForm` as a component and add it to the competencies page
- Allow users to add new competencies from the form which would be displayed as a list on the competencies page

Resolves issue #275 

image

## Checklist


EnriqueTMT (approved):

webapp/src/components/CompetenciesPage.tsx
```diff
@@ -1,13 +1,17 @@
...
+import { EntityId, Competency as APICompetency } from '../types/api';
 import useApiData from '../helpers/useApiData';
 
 const CompetenciesPage = () => {
+  const [newlyCreatedCompetencyIds, setNewlyCreatedCompetencyIds] = useState<
+    EntityId[]
+  >([]);
   const { data: competencies, error } = useApiData<APICompetency[]>({
-    deps: [],
+    deps: [newlyCreatedCompetencyIds],
```
pthan1, line 14: Can you tell me what is happening here?  Are the returned result of the POST requests added into the deps array? Also, what's the mechanism for displaying the new competency on the page?

webapp/src/components/CreateCompetencyForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+      });
+  };
+  return (
+    <form onSubmit={onSubmit}>
+      <Card>
+        <CardContent sx={{ pb: 0 }}>
+          <h2>New Competency</h2>
+        </CardContent>
+        <CardContent sx={{ pt: 0 }}>
+          <TextField
```
pthan1, line 60: There's a maximum character limit of 255 characters that was defined in the database migration.  Would it be valuable to have some sort of character limit defined in the front-end to prevent users from trying to submit labels that are longer than 255 characters?

pthan1 (approved):
@eileenxue Both of the fields are required.  I cannot leave the 'Title' field blank and input something into the 'Description' field and vice versa.  Would it be valuable to have any of the fields not required?  There were previous discussions on being able for users to create competencies without having a description.

I fetched down the changes and don't see any errors.

pthan1:
When I try to submit a competency without title nor description, only one error message shows on the 'Title' input saying that that field is required.  I think it would be valuable to display error messages for both fields at the same time.  

I believe that UX would be better because it will tell users to input something into both fields as opposed to one at a time.  For example, this would avoid user frustration if they only input a title, only to submit and have another error telling them that they need to input a description.

webapp/src/components/CompetenciesPage.tsx
eileenxue: When I added `newlyCreatedCompetencyIds` into the array, it was able to populate new competencies dynamically on the competencies page. Otherwise, you would need to refresh the page to see the changes.

BRashad:
on my local machine, all work as per the ticket requirements. Subject to question feedback on server-side error upon unsuccessful competency insert into DB. the case is good to be merged. 

Screen Shot 2022-02-08 at 2 29 12 PM
Screen Shot 2022-02-08 at 2 29 26 PM

BRashad:
I have only one question: Have you tested simulated the server-side error in case the user competency input can not be successfully inserted into DB ?

webapp/src/components/CompetenciesPage.tsx
pthan1: How does the app know what the competency's label and description is?

LouisDG6:
> I have only one question: Have you tested simulated the server-side error in case the user competency input can not be successfully inserted into DB ?

Yes, its been tested.

webapp/src/components/CreateCompetencyForm.tsx
eileenxue: I think showing them a character limit would be a nice to have feature in the future, for this form and other form fields on the application. For now, when I tried to submit a new competency into the form with a title that's longer than 255 characters, it would not go through and there's a general message that shows up. I believe that will satisfy our issue ticket. 

image

webapp/src/components/CompetenciesPage.tsx
eileenxue: There's a state for newly created competencies on the create competency form on the competency page. So I guess it sets the state to the new array of existing competencies plus any new ones that gets added after that.

pthan1: Thanks!  I think the newlyCreatedComptencies array only holds the ids of the competencies.  The useApiData function on line 13 has a a fetch and returns the list of competencies and stores it in the data property as competencies.

webapp/src/components/CreateCompetencyForm.tsx
pthan1: Thanks for following up!  That's great!  Yes, it'll be good to have some sort of feature in the future, that'll notify the user why it wasn't saved.

PR #299, opened 2022-02-08
Run migrations in a container as a service in production

davidvandusen:
## Proposed changes

Build a flyway image on release, and run `flyway migrate` as a service when the production stack is deployed.

This is a big step toward #90.

## Checklist

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -2,9 +2,6 @@ version: "3.7"
 
 services:
   api:
-    depends_on:
-      - flyway
-      - redis
     environment:
       - API_HOST=0.0.0.0
       - GITHUB_CLIENT_ID
@@ -26,9 +23,6 @@ services:
     secrets:
       - FLYWAY_CONFIG
   nginx:
-    depends_on:
-      - api
-      - webapp
     image: ghcr.io/opentree-education/rhizone-lms_nginx:latest
     ports:
       - "80:80"
@@ -38,6 +32,4 @@ services:
   redis:
     image: redis:6.2-alpine
   webapp:
-    depends_on:
-      - api
     image: ghcr.io/opentree-education/rhizone-lms_webapp:latest
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -3,6 +3,7 @@ version: "3.7"
 services:
   api:
     depends_on:
+      - flyway
       - redis
     environment:
       - API_HOST=0.0.0.0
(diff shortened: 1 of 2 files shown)


docker-compose.production.yml
```diff
@@ -17,6 +18,13 @@ services:
       - SESSION_SECRET
       - WEBAPP_ORIGIN=https://rhi.zone
     image: ghcr.io/opentree-education/rhizone-lms_api:latest
+  flyway:
+    command: -configFiles=/run/secrets/FLYWAY_CONFIG migrate
+    deploy:
+      replicas: 1
+    image: ghcr.io/opentree-education/rhizone-lms_flyway:latest
+    secrets:
+      - FLYWAY_CONFIG
```
davidvandusen, line 27: The flyway config will be managed as a file:

https://flywaydb.org/documentation/configuration/configfile.html

Using Docker Swarm Secrets:

https://docs.docker.com/engine/swarm/secrets/

docker-compose.production.yml
```diff
@@ -30,4 +38,6 @@ services:
   redis:
     image: redis:6.2-alpine
   webapp:
+    depends_on:
+      - api
```
pthan1, line 42: Why did the webapp not have a dependency on the api service before?

pthan1 (approved):

docker-compose.production.yml
davidvandusen: To be fair, since this compose file is being deployed as a Stack in a Swarm, the `depends_on` config will be ignored. Maybe all the `depends_on` should just be removed since they don't do anything.

While I was updating the file, I put it in there out of habit, since this compose file can be use with Docker Compose as well as Docker Swarm. That's speculative generalization, though.

pthan1: Would it be valuable/necessary to find a solution that'll allow any dependencies to be expressed?

davidvandusen: The lack of reliability caused by not defining version and service dependencies is within our tolerance. There are some situations that we can avoid by having app code check versions and availability of deployed dependencies, but the app isn't used enough to make that work worth the effort.

pthan1: Thanks for explaining.  

When you say "not defining version" are you referring to Docker compose version property defined in a Docker Compose file and that it is only "informative" and is deprecated? Are are you referring to versioning of the app itself?

What usage metrics would be useful to monitor in order to measure when we would need to have app code check versions and availability of deployed dependencies?

pthan1: Thanks for the explanation!

EnriqueTMT (approved):
Doesn't look like it will crash production. Thumbs up for me. I approve

PR #301, opened 2022-02-09
Remove flyway service in prod stack because it keeps restarting

davidvandusen:
## Proposed changes

When deploying the production stack to the swarm, the flyway service continuously restarts even though it exits with `0`. Since we only want the migrations to run once per deploy, this is undesirable. Instead, flyway will have to be run as a one-off elsewhere in the pipeline. Until then, removing it from the stack config so we can deploy other stuff.

One step back in #90.

## Checklist

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -15,13 +15,6 @@ services:
       - SESSION_SECRET
       - WEBAPP_ORIGIN=https://rhi.zone
     image: ghcr.io/opentree-education/rhizone-lms_api:latest
-  flyway:
-    command: -configFiles=/run/secrets/FLYWAY_CONFIG migrate
-    deploy:
-      replicas: 1
-    image: ghcr.io/opentree-education/rhizone-lms_flyway:latest
-    secrets:
-      - FLYWAY_CONFIG
   nginx:
     image: ghcr.io/opentree-education/rhizone-lms_nginx:latest
     ports:


BRashad (approved):

pthan1 (commented):
PR #299 adds jobs to `.github/workflows/build_and_publish_images.yml`file.  Do those changes need to be removed as well?

davidvandusen:
@pthan1 We still need to build and push the flyway image, so that stuff stays. This just removes the service from the stack, but the image will be run elsewhere.

pthan1:
> @pthan1 We still need to build and push the flyway image, so that stuff stays. This just removes the service from the stack, but the image will be run elsewhere.

Ok thanks for clarifying!  Do you have an idea of where it'll be run?  If so, where?

davidvandusen:
@pthan1 There are plenty of options for where to run it. The idea I was trying out—running migrations on the swarm—was in an attempt to centralize secret management to Docker Secrets, which only work in swarm mode. It's still an option to create a new service for each migration run with `--restart-condition=none` to just run the migrations once, but I'm leaning away from running migrations as a service.

That is, I'm starting to lean toward using GitHub Secrets just for migrations. Then they could be run with `docker run` and block the rest of the deployment on failure.

pthan1:
> @pthan1 There are plenty of options for where to run it. The idea I was trying out—running migrations on the swarm—was in an attempt to centralize secret management to Docker Secrets, which only work in swarm mode. It's still an option to create a new service for each migration run with `--restart-condition=none` to just run the migrations once, but I'm leaning away from running migrations as a service.
> 
> That is, I'm starting to lean toward using GitHub Secrets just for migrations. Then they could be run with `docker run` and block the rest of the deployment on failure.

Thanks

pthan1 (approved):

Issue #277, opened 2022-01-21
There is no API endpoint for updating competencies


davidvandusen:
## Expected behaviour

Sending an HTTP `PUT` request to `/competencies/:id` updates the specified competency if the current user is the one who originally authored the competency.

## Actual behaviour

There is no way to update a competency via the API.

## Details and resources

The same validations that take place in the `POST` endpoint apply to the `PUT` endpoint.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #302, opened 2022-02-09
Add api endpoint for updating competencies

BRashad:
## Proposed changes

This is partial resolution for ticket #277.
updateCompetency function and test for querying a specific competency for a user who authored it.

## Checklist

### DIFF
--- a/api/src/middleware/__tests__/competenciesRouter.ts
+++ b/api/src/middleware/__tests__/competenciesRouter.ts
@@ -1,8 +1,10 @@
 import { collectionEnvelope, itemEnvelope } from '../responseEnvelope';
 import {
+  authorizeCompetencyUpdate,
   countCompetencies,
   listCompetencies,
   createCompetency,
+  updateCompetency,
 } from '../../services/competenciesService';
 import { createAppAgentForRouter, mockPrincipalId } from '../routerTestUtils';
 import competenciesRouter from '../competenciesRouter';
@@ -11,6 +13,8 @@ jest.mock('../../services/competenciesService.ts');
 const mockCountCompetencies = jest.mocked(countCompetencies);
 const mockListCompetencies = jest.mocked(listCompetencies);
 const mockCreateCompetency = jest.mocked(createCompetency);
+const mockUpdateCompetency = jest.mocked(updateCompetency);
+const mockAuthorizeCompetencyUpdate = jest.mocked(authorizeCompetencyUpdate);
 
 describe('competenciesRouter', () => {
   const appAgent = createAppAgentForRouter(competenciesRouter);
@@ -62,6 +66,7 @@ describe('competenciesRouter', () => {
       const competencyId = 4;
       const competency = { id: competencyId };
       mockPrincipalId(principalId);
+
       mockCreateCompetency.mockResolvedValue(competency);
       appAgent
         .post('/')
@@ -97,6 +102,82 @@ describe('competenciesRouter', () => {
       mockCreateCompetency.mockRejectedValue(new Error());
       appAgent
         .post('/')
+        .send({
+          label: 'test',
+          description: 'test',
+        })
+        .expect(500, done);
+    });
+  });
+
+  describe('PUT /:id', () => {
+    it('should update a competency of user who authored it', done => {
+      const principalId = 2;
+      const competencyId = 3;
+      const label = 'label';
+      const description = 'description';
+      const competency = {
+        id: competencyId,
+      };
+      mockPrincipalId(principalId);
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      mockUpdateCompetency.mockResolvedValue();
+      appAgent
+        .put('/3')
+        .send({
+          label: label,
+          description: description,
+        })
+        .expect(200, itemEnvelope(competency), err => {
+          expect(mockAuthorizeCompetencyUpdate).toHaveBeenCalledWith(
+            principalId,
+            competencyId
+          );
+          expect(mockUpdateCompetency).toHaveBeenCalledWith(
+            competencyId,
+            label,
+            description
+          );
+          done(err);
+        });
+    });
+
+    it('should respond with a validation error if given id is less than 1', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      appAgent.put('/0').send({ label: '', description: '' }).expect(400, done);
+    });
+
+    it('should respond with an internal server error if an error was thrown while authorising a competency update', done => {
+      mockAuthorizeCompetencyUpdate.mockRejectedValue(new Error());
+      appAgent.put('/3').send({ label: '', description: '' }).expect(500, done);
+    });
+
+    it('should respond with a bad request error if user can not be authorised', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(false);
+      appAgent.put('/3').send({ label: '', description: '' }).expect(404, done);
+    });
+
+    it('should respond with a validation error if label is not a string', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      appAgent
+        .put('/3')
+        .send({ label: null, description: '' })
+        .expect(422, done);
+    });
+
+    it('should respond with a validation error if description is not a string', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      appAgent
+        .put('/3')
+        .send({ label: '', description: null })
+        .expect(422, done);
+    });
+
+    it('should respond with an internal server error if an error is thrown when updating the competency', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      mockUpdateCompetency.mockRejectedValue(new Error());
+      appAgent
+        .put('/3')
         .send({
           label: 'test',
           description: 'test',
--- a/api/src/middleware/__tests__/competenciesRouter.ts
+++ b/api/src/middleware/__tests__/competenciesRouter.ts
@@ -10,11 +10,11 @@ import { createAppAgentForRouter, mockPrincipalId } from '../routerTestUtils';
 import competenciesRouter from '../competenciesRouter';
 
 jest.mock('../../services/competenciesService.ts');
-const mockCountCompetencies = jest.mocked(countCompetencies);
-const mockListCompetencies = jest.mocked(listCompetencies);
-const mockCreateCompetency = jest.mocked(createCompetency);
-const mockUpdateCompetency = jest.mocked(updateCompetency);
 const mockAuthorizeCompetencyUpdate = jest.mocked(authorizeCompetencyUpdate);
+const mockCountCompetencies = jest.mocked(countCompetencies);
+const mockCreateCompetency = jest.mocked(createCompetency);
+const mockListCompetencies = jest.mocked(listCompetencies);
+const mockUpdateCompetency = jest.mocked(updateCompetency);
 
 describe('competenciesRouter', () => {
   const appAgent = createAppAgentForRouter(competenciesRouter);
@@ -116,19 +116,15 @@ describe('competenciesRouter', () => {
       const competencyId = 3;
       const label = 'label';
       const description = 'description';
-      const competency = {
-        id: competencyId,
-      };
       mockPrincipalId(principalId);
       mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
-      mockUpdateCompetency.mockResolvedValue();
       appAgent
         .put('/3')
         .send({
           label: label,
           description: description,
         })
-        .expect(200, itemEnvelope(competency), err => {
+        .expect(200, err => {
           expect(mockAuthorizeCompetencyUpdate).toHaveBeenCalledWith(
             principalId,
             competencyId
@@ -142,17 +138,12 @@ describe('competenciesRouter', () => {
         });
     });
 
-    it('should respond with a validation error if given id is less than 1', done => {
+    it('should respond with a bad request error if the given id is less than 1', done => {
       mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
       appAgent.put('/0').send({ label: '', description: '' }).expect(400, done);
     });
 
-    it('should respond with an internal server error if an error was thrown while authorising a competency update', done => {
-      mockAuthorizeCompetencyUpdate.mockRejectedValue(new Error());
-      appAgent.put('/3').send({ label: '', description: '' }).expect(500, done);
-    });
-
-    it('should respond with a bad request error if user can not be authorised', done => {
+    it('should respond with not found if user can not be authorised', done => {
       mockAuthorizeCompetencyUpdate.mockResolvedValue(false);
       appAgent.put('/3').send({ label: '', description: '' }).expect(404, done);
     });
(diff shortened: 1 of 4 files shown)


api/src/services/competenciesService.ts
```diff
@@ -34,3 +34,11 @@ export const createCompetency = async (
   });
   return { id: competencyId };
 };
+
+export const updateCompetency = async (id: number, principalId: number) => {
+  const competency = await db('competencies')
+    .select('description', 'id', 'label', 'principal_id')
+    .where({ id: id, principal_id: principalId });
+
```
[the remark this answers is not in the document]
BRashad: Can you please elaborate on which update are you referring to?

pthan1 (changes requested):
Hey Rashad, the service function is only querying for a matching competency and returns the competency.  it doesn't perform any sort of update on that existing record (if found).  I would try writing a working test and seeing if it passes!

api/src/services/competenciesService.ts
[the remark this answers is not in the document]
pthan1: By "update" I mean updating a label or description of a competency that already exists in the database.

BRashad:
> 

Yes this is the 2nd part of the ticket which i will push later.

api/src/services/competenciesService.ts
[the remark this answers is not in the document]
BRashad: Yes this is the 2nd part of the ticket i will push later.

pthan1:
> 

OK, this is a great start!  I know you follow TDD so I suggest re-writing the test to sufficiently test what you'd like the function to do and then push up the remaining code.

EnriqueTMT (approved):
Everything looks solid. I'll approve it!

api/src/middleware/competenciesRouter.ts
```diff
@@ -48,4 +50,47 @@ competenciesRouter.post('/', async (req, res, next) => {
   res.status(201).json(itemEnvelope(competency));
 });
 
+competenciesRouter.put('/:id', async (req, res, next) => {
+  const { principalId } = req.session;
+  const { id } = req.params;
+
```
davidvandusen, line 56: 👾

api/src/middleware/competenciesRouter.ts
```diff
@@ -48,4 +50,47 @@ competenciesRouter.post('/', async (req, res, next) => {
...
+  }
+  if (!isAuthorized) {
+    next(
+      new NotFoundError(
+        `A competency with the id "${competencyId}" could not be found.`
+      )
+    );
+    return;
+  }
+  const { label, description } = req.body;
```
pthan1, line 78: I think this declaration should be at the top of the function.

davidvandusen: Now that we have blocked scoped variable declarations and we don't have to worry about hoisting, we can define variables as close to their first use as possible in order to make the code more readable and maintainable.

BRashad:
@pthan1 Please review the latest commit. thanks in advance.

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -104,4 +109,80 @@ describe('competenciesRouter', () => {
...
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      appAgent.put('/0').send({ label: '', description: '' }).expect(400, done);
+    });
+
+    it('should respond with an internal server error if an error was thrown while authorising a competency update', done => {
+      mockAuthorizeCompetencyUpdate.mockRejectedValue(new Error());
+      appAgent.put('/3').send({ label: '', description: '' }).expect(500, done);
+    });
+
+    it('should respond with a bad request error if user can not be authorised', done => {
```
davidvandusen, line 155: This description says "bad request error" but the code tests for `404`, a not found error. Looks like the description of the test should be updated.

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -104,4 +109,80 @@ describe('competenciesRouter', () => {
...
+          expect(mockUpdateCompetency).toHaveBeenCalledWith(
+            competencyId,
+            label,
+            description
+          );
+          done(err);
+        });
+    });
+
+    it('should respond with a validation error if given id is less than 1', done => {
```
davidvandusen, line 145: The description says "validation error" but the code tests for `400`, a bad request error. Looks like the description of the test should be updated.

api/src/middleware/competenciesRouter.ts
```diff
@@ -48,4 +50,47 @@ competenciesRouter.post('/', async (req, res, next) => {
...
+    next(new ValidationError('description must be a string'));
+    return;
+  }
+  try {
+    await updateCompetency(competencyId, label, description);
+  } catch (error) {
+    next(error);
+    return;
+  }
+  res.status(200).json(itemEnvelope({ id: competencyId }));
```
davidvandusen, line 93: The default status when json is sent is `200`. Is it necessary to call `.status(200)` here?

api/src/services/__tests__/competenciesService.ts
```diff
@@ -62,4 +64,28 @@ describe('competenciesService', () => {
       });
     });
   });
+
+  describe('updateCompetency', () => {
+    it('should update the specified competency if the current user is the one who originally authored the competency', async () => {
```
davidvandusen, line 69: This test description says "if the current user is the one who originally authored the competency" but there is no code in the test that asserts anything about the "current user". Looks like the test description needs to be updated.

api/src/services/__tests__/competenciesService.ts
```diff
@@ -62,4 +64,28 @@ describe('competenciesService', () => {
...
+      const label = 'label';
+      const description = 'description';
+      mockQuery(
+        'update `competencies` set `label` = ?, `description` = ? where `id` = ?',
+        [label, description, id],
+        1
+      );
+      await updateCompetency(id, label, description);
+    });
+    it('should authorizeCompetencyUpdate for a user who authored the competency', async () => {
```
davidvandusen, line 80: This example is testing the `authorizeCompetencyUpdate` function, but it's inside of the `describe('updateCompetency'` block. It looks like a new block should be created for testing the function.

Also, using the function name in the test description makes it hard to understand what the test is meant to assert. I think the test description should be updated to explain what's being asserted in the test itself.

api/src/services/__tests__/competenciesService.ts
```diff
@@ -62,4 +64,28 @@ describe('competenciesService', () => {
...
+      const principalId = 2;
+      const competencyId = 3;
+      mockQuery(
+        'select `id` from `competencies` where `id` = ? and `principal_id` = ?',
+        [competencyId, principalId],
+        [competencyId, principalId]
```
davidvandusen, line 86: The mocked response from this query doesn't appear to conform to what knex would be returning from this query. It's really important when mocking apis to ensure that the mocks behave exactly like the original apis.

davidvandusen (commented):
Looks like some of the test descriptions need to be updated to reflect the content of the tests.

BRashad:
I've gone through the comment. I will make changes as suggested and push them.

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: Will update to `'should respond with competency not found if the user can not be authorised'`

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: Will update with `'should respond with a bad request error if the given id is less than 1'` message.

api/src/middleware/competenciesRouter.ts
BRashad: yes indeed. simplified to `res.json(...)`

api/src/services/__tests__/competenciesService.ts
BRashad: Updated test description to `'should update the specified/selected competency'`

api/src/services/__tests__/competenciesService.ts
BRashad: I will separate the test block for the authorization of the competency update.  I will make changes to the test description as suggested.

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -11,6 +13,8 @@ jest.mock('../../services/competenciesService.ts');
 const mockCountCompetencies = jest.mocked(countCompetencies);
 const mockListCompetencies = jest.mocked(listCompetencies);
 const mockCreateCompetency = jest.mocked(createCompetency);
+const mockUpdateCompetency = jest.mocked(updateCompetency);
+const mockAuthorizeCompetencyUpdate = jest.mocked(authorizeCompetencyUpdate);
```
pthan1, line 17: Should these variables also be listed alphabetically like their imports?

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -104,4 +109,80 @@ describe('competenciesRouter', () => {
...
+      const principalId = 2;
+      const competencyId = 3;
+      const label = 'label';
+      const description = 'description';
+      const competency = {
+        id: competencyId,
+      };
+      mockPrincipalId(principalId);
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      mockUpdateCompetency.mockResolvedValue();
```
pthan1, line 124: The updateCompetency function doesn't return any values.  Do we need to mock a resolved value in this test?

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -104,4 +109,80 @@ describe('competenciesRouter', () => {
...
     });
   });
+
+  describe('PUT /:id', () => {
+    it('should update a competency of user who authored it', done => {
+      const principalId = 2;
+      const competencyId = 3;
+      const label = 'label';
+      const description = 'description';
+      const competency = {
```
pthan1, line 119: This `competency` variable is declared but never used.  is it necessary in this test?

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -104,4 +109,80 @@ describe('competenciesRouter', () => {
...
+          done(err);
+        });
+    });
+
+    it('should respond with a validation error if given id is less than 1', done => {
+      mockAuthorizeCompetencyUpdate.mockResolvedValue(true);
+      appAgent.put('/0').send({ label: '', description: '' }).expect(400, done);
+    });
+
+    it('should respond with an internal server error if an error was thrown while authorising a competency update', done => {
```
pthan1, line 150: Convention has the `internal server error` tests at the bottom of the `describe` blocks.  I think this test could be moved to the bottom as well.

api/src/middleware/__tests__/competenciesRouter.ts
pthan1: To follow convention, I think we could keep the test description but replace `bad request` with `not found`.  See[ ](https://github.com/OpenTree-Education/rhizone-lms/blob/main/api/src/middleware/__tests__/meetingsRouter.ts#L71)

api/src/middleware/competenciesRouter.ts
pthan1: Thanks for the explanation.

api/src/middleware/__tests__/competenciesRouter.ts
davidvandusen: It is used on line 131. That said, when a variable is only used once, a lot of times it's more readable to just inline the value, unless it's unclear what the value represents. Luckily, we have a lint rule that fails if a variable is declared and actually never used.

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: Amended to alphabetical order as suggested.

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: Based on the discussion on slack i will remove this mock function call

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: This is meant to be used for testing update competency function. I will remove it since there is no need to test void functions.

api/src/services/__tests__/competenciesService.ts
BRashad: I made changes in the test matching the knex call output.

api/src/middleware/__tests__/competenciesRouter.ts
BRashad: moved the test as suggested to the bottom.

api/src/services/__tests__/competenciesService.ts
```diff
@@ -62,4 +64,31 @@ describe('competenciesService', () => {
...
+  });
+
+  describe('authorizeCompetencyUpdate', () => {
+    it('should authorize the specific competency update ', async () => {
+      const principalId = 2;
+      const competencyId = 3;
+      mockQuery(
+        'select `id` from `competencies` where `id` = ? and `principal_id` = ?',
+        [competencyId, principalId],
+        [competencyId]
```
davidvandusen, line 89: I just took a look at the knex docs for `update` and it looks like it could return a number of different things, but just an array with the `id` of the updated record isn't one of the options.

https://knexjs.org/#Builder-update

Can you confirm that this matches with the actual behaviour when the app is running? If it's different, this test could be confusing or may permit regressions.

BRashad: The mock query output now matches the authorized competency update function output.

davidvandusen (approved):
Nice work! I think this is good to go.

api/src/middleware/__tests__/competenciesRouter.ts
```diff
@@ -1,16 +1,20 @@
 import { collectionEnvelope, itemEnvelope } from '../responseEnvelope';
 import {
+  authorizeCompetencyUpdate,
   countCompetencies,
   listCompetencies,
   createCompetency,
+  updateCompetency,
 } from '../../services/competenciesService';
 import { createAppAgentForRouter, mockPrincipalId } from '../routerTestUtils';
 import competenciesRouter from '../competenciesRouter';
```
pthan1, line 10: These imports should be in alphabetical order.

api/src/middleware/competenciesRouter.ts
```diff
@@ -1,11 +1,13 @@
 import { Router } from 'express';
 
-import { ValidationError } from './httpErrors';
+import { BadRequestError, NotFoundError, ValidationError } from './httpErrors';
```
pthan1, line 3: These imports should be in alphabetical order.

api/src/middleware/competenciesRouter.ts
pthan1, line 56: I think this is an extra blank line.

api/src/middleware/competenciesRouter.ts
```diff
@@ -48,4 +50,47 @@ competenciesRouter.post('/', async (req, res, next) => {
...
+competenciesRouter.put('/:id', async (req, res, next) => {
+  const { principalId } = req.session;
+  const { id } = req.params;
+
+  const competencyId = Number(id);
+  if (!Number.isInteger(competencyId) || competencyId < 1) {
+    next(new BadRequestError(`"${id}" is not a valid competency id.`));
+    return;
+  }
+
```
pthan1, line 62: I think this is an extra blank line.

pthan1 (approved):
The second half of the PR is finished!  Great job refactoring the tests.

PR #305, opened 2022-02-10
Allow principal authors to edit competencies

pthan1:
## Proposed changes

Refactors the Create Competency Form in order to allow users to edit competencies they've authored.
Adds UI changes in order to populate edit form.

## Checklist

Screen Shot 2022-02-10 at 2 32 43 PM

### DIFF
--- a/webapp/src/components/CompetenciesPage.tsx
+++ b/webapp/src/components/CompetenciesPage.tsx
@@ -7,11 +7,11 @@ import { EntityId, Competency as APICompetency } from '../types/api';
 import useApiData from '../helpers/useApiData';
 
 const CompetenciesPage = () => {
-  const [newlyCreatedCompetencyIds, setNewlyCreatedCompetencyIds] = useState<
+  const [changedCompetencyIds, setChangedCompetencyIds] = useState<
     EntityId[]
   >([]);
   const { data: competencies, error } = useApiData<APICompetency[]>({
-    deps: [newlyCreatedCompetencyIds],
+    deps: [changedCompetencyIds],
     path: `/competencies`,
     sendCredentials: true,
   });
@@ -28,15 +28,15 @@ const CompetenciesPage = () => {
         <Grid item xs={12} md={8}>
           <Box>
             <CreateCompetencyForm
-              onCompetencyCreated={id =>
-                setNewlyCreatedCompetencyIds([...newlyCreatedCompetencyIds, id])
-              }
+              onCompetenciesChanged={id =>
+                setChangedCompetencyIds([...changedCompetencyIds, id])}
             />
           </Box>
           <Box my={6}>
             {competencies.length === 0 && <p>There are no competencies.</p>}
             {competencies.map(({ id, label, description, principal_id }) => (
-              <Competency key={id} description={description} label={label} principal_id={principal_id} />
+              <Competency key={id} id={id} description={description} label={label} principal_id={principal_id} onCompetenciesChanged={id =>
+                setChangedCompetencyIds([...changedCompetencyIds, id])} />
             ))}
           </Box>
         </Grid>
--- a/webapp/src/components/CompetenciesPage.tsx
+++ b/webapp/src/components/CompetenciesPage.tsx
@@ -28,14 +28,14 @@ const CompetenciesPage = () => {
         <Grid item xs={12} md={8}>
           <Box>
             <CreateCompetencyForm
-              onCompetenciesChanged={id =>
+              onCompetencyChanged={id =>
                 setChangedCompetencyIds([...changedCompetencyIds, id])}
             />
           </Box>
           <Box my={6}>
             {competencies.length === 0 && <p>There are no competencies.</p>}
-            {competencies.map(({ id, label, description, principal_id }) => (
-              <Competency key={id} id={id} description={description} label={label} principal_id={principal_id} onCompetenciesChanged={id =>
+            {competencies.map(({ id, label, description, principal_id: principalId }) => (
+              <Competency key={id} id={id} description={description} label={label} principalId={principalId} onCompetencyChanged={id =>
                 setChangedCompetencyIds([...changedCompetencyIds, id])} />
             ))}
           </Box>
--- a/webapp/src/components/CompetenciesPage.tsx
+++ b/webapp/src/components/CompetenciesPage.tsx
@@ -2,7 +2,7 @@ import { Box, Container, Grid } from '@mui/material';
 import React, { useState } from 'react';
 
 import Competency from './Competency';
-import CreateCompetencyForm from './CreateCompetencyForm';
+import CreateOrUpdateCompetencyForm from './CreateOrUpdateCompetencyForm';
 import { EntityId, Competency as APICompetency } from '../types/api';
 import useApiData from '../helpers/useApiData';
 
@@ -27,7 +27,7 @@ const CompetenciesPage = () => {
       <Grid container justifyContent="center">
         <Grid item xs={12} md={8}>
           <Box>
-            <CreateCompetencyForm
+            <CreateOrUpdateCompetencyForm
               onCompetencyChanged={id =>
                 setChangedCompetencyIds([...changedCompetencyIds, id])
               }
--- a/webapp/src/components/CompetenciesPage.tsx
+++ b/webapp/src/components/CompetenciesPage.tsx
@@ -7,9 +7,9 @@ import { EntityId, Competency as APICompetency } from '../types/api';
 import useApiData from '../helpers/useApiData';
 
 const CompetenciesPage = () => {
-  const [changedCompetencyIds, setChangedCompetencyIds] = useState<
-    EntityId[]
-  >([]);
+  const [changedCompetencyIds, setChangedCompetencyIds] = useState<EntityId[]>(
+    []
+  );
   const { data: competencies, error } = useApiData<APICompetency[]>({
     deps: [changedCompetencyIds],
     path: `/competencies`,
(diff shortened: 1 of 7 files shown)


webapp/src/components/CompetenciesPage.tsx
```diff
@@ -28,15 +28,15 @@ const CompetenciesPage = () => {
         <Grid item xs={12} md={8}>
           <Box>
             <CreateCompetencyForm
-              onCompetencyCreated={id =>
-                setNewlyCreatedCompetencyIds([...newlyCreatedCompetencyIds, id])
-              }
+              onCompetenciesChanged={id =>
```
davidvandusen, line 31: Interesting that `Competencies` here is plural, even though this form only changes one competency at a time.

webapp/src/components/CompetenciesPage.tsx
```diff
@@ -28,15 +28,15 @@ const CompetenciesPage = () => {
...
+                setChangedCompetencyIds([...changedCompetencyIds, id])}
             />
           </Box>
           <Box my={6}>
             {competencies.length === 0 && <p>There are no competencies.</p>}
-            {competencies.map(({ id, label, description }) => (
-              <Competency key={id} description={description} label={label} />
+            {competencies.map(({ id, label, description, principal_id }) => (
```
davidvandusen, line 37: Variables in JavaScript should be camel case. `principal_id` ought to be aliased to `principalId` then.

webapp/src/components/CompetenciesPage.tsx
```diff
@@ -28,15 +28,15 @@ const CompetenciesPage = () => {
...
+              <Competency key={id} id={id} description={description} label={label} principal_id={principal_id} onCompetenciesChanged={id =>
```
davidvandusen, line 38: Props in React should be camel case. `principal_id` should be `principalId` in that case.

webapp/src/components/Competency.tsx
```diff
@@ -1,15 +1,32 @@
-import React from 'react';
+import { Button } from '@mui/material';
+import React, { useContext, useState } from 'react';
+import CreateCompetencyForm from './CreateCompetencyForm';
+import SessionContext from './SessionContext';
+
```
davidvandusen, line 5: The blank line in the imports section is meant to come in between `node_modules` imports and relative imports.

webapp/src/components/Competency.tsx
```diff
@@ -1,15 +1,32 @@
...
 }
 
-const Competency = ({ label, description }: CompetencyProps) => (
-  <div>
-    <h2>{label}</h2>
-    <p>{description}</p>
-  </div>
-);
+const Competency = ({ description, id, label, onCompetenciesChanged, principal_id }: CompetencyProps) => {
+  const { principalId } = useContext(SessionContext);
```
davidvandusen, line 17: `principalId` here would shadow the `principalId` prop. It might be worth aliasing this to `sessionPrincipalId` or something like that in order to differentiate it.

webapp/src/components/Competency.tsx
```diff
@@ -1,15 +1,32 @@
...
+  const [toggleForm, setToggleForm] = useState(false);
+
+  return (
+    <div>
+      <h2>{label}</h2>
+      <p>{description}</p>
+      {principal_id === principalId && <Button onClick={() => setToggleForm(true)}>Edit Competency</Button>}
```
davidvandusen, line 24: If this is a toggle button, shouldn't it set `toggleForm` to `!toggleForm` instead of `true`?

webapp/src/components/CreateCompetencyForm.tsx
```diff
@@ -54,7 +62,7 @@ const CreateCompetencyForm = ({
     <form onSubmit={onSubmit}>
       <Card>
         <CardContent sx={{ pb: 0 }}>
-          <h2>New Competency</h2>
+          {competencyId ? <h3>Edit Competency</h3> : <h2>New Competency</h2>}
```
davidvandusen, line 65: Why is it sometimes an `<h3>` and sometimes an `<h2>`?

eileenxue:
@pthan1 FYI: A recent PR has created some conflicts with the competenciesService files on this branch.

webapp/src/components/CreateCompetencyForm.tsx
pthan1: The "Edit Competency" header falls below a competency's which is a `h2'.  We might need to refactor the structure of the HTML.

webapp/src/components/Competency.tsx
pthan1: Yes, it should.

webapp/src/components/Competency.tsx
pthan1: Thanks.  I'll move the blank line.

webapp/src/components/Competency.tsx
pthan1: Thanks, `sessionPrincipalId` would make a lot more sense.  To be sure, do you mean that this name would shadow the `principal_id` prop declared on line 13?  Thanks.

webapp/src/components/CompetenciesPage.tsx
pthan1: It should be singular.  I'll make that change.

webapp/src/components/CompetenciesPage.tsx
pthan1: I'll make this change along with the one below.

webapp/src/components/CreateCompetencyForm.tsx
davidvandusen: I guess that's fine for now. React really makes it hard to have semantic hierarchy, because you can compose components however you want. Without taking the hierarchy level as a prop, it's not really possible to know what level of heading to use. In this case, it's probably fine to just assume like this, and we can refactor it later on if it ends up mattering.

eileenxue: Originally for the form hierarchy and heading types, we followed the reflection's form and how they used a card with an h2 heading. As for the individual competency, that component has not included the MUI styles yet, but I would argue those should be h3 or lower. Since Create/Edit Competency is the same form, should it not be h2? Also, maybe the way I envisioned that was that the Create/Edit Form would always be near the top of the UI...OR the only way it would make sense would be to directly edit the individual competency text values (without the additional edit form).

eileenxue:
Right now the behaviour is that when you click the `Edit Competency` button, a new form appears. Would it be possible to show that without duplication of the existing competency that you're trying to edit? Not sure how, but maybe the easiest way would be to hide it and after saving it, it will populate with the new data. It just makes the UI less cluttered.
image

webapp/src/components/CreateCompetencyForm.tsx
eileenxue: I made a newer comment at the end of the PR, but I figure here might be a better place to show the screenshot of what I mean. Instead of showing duplicated data, would it be better to just show it once in the form? Not sure if that'll change your entire implementation. Or just hide the competency info (top half) until new changes are saved or the user cancels out of the form.

image

eileenxue:
So currently when I click `Edit Competency` and then I changed my mind, there's no way for me to exit the form. maybe a secondary `Cancel` button with MUI should do the trick? 

Screen Shot 2022-02-15 at 4 11 53 PM

webapp/src/components/CreateCompetencyForm.tsx
pthan1: @eileenxue Re: the value of forms - Yes, I think it could be better if we show the data once in the form.  I think improving the UX should be handled in another issue/ticket where a proof of concept could be explored further.
One question I want to throw out there is what if user's wanted to compare their new changes with the old data?  

- We are using the `PUT` method to update competencies.  A `PUT` method replaces a record and all of the columns of data.  So, we need to populate the form with the existing label/description so that we can submit all of the data needed for the `PUT`, or else a column would be replaced by an empty string.

Re: heading levels - If we are on the `Competencies` page, shouldn't the competencies have a higher heading level than the `Create Competency` and `Edit Competency` headings.  My line of thinking was that a screen reader should list the Competencies before "Edit Competency."  Also, the Edit form doesn't show up until you click on the toggle button and even then, I think it should have a lower heading level than a competency.  Maybe we could move the competencies and the `Edit Competency` headings one level down?

pthan1:
> So currently when I click `Edit Competency` and then I changed my mind, there's no way for me to exit the form. maybe a secondary `Cancel` button with MUI should do the trick?
> 
> Screen Shot 2022-02-15 at 4 11 53 PM

This is a great idea.  Should we set up a separate ticket for this?

webapp/src/components/Competency.tsx
eileenxue: @pthan1 Is this still there in the latest commit?

eileenxue (changes requested):

webapp/src/components/Competency.tsx
eileenxue: Never mind, I noticed David resolved the convo during my review process.

eileenxue (approved):
I think it's good to go, you seem to have resolved the previous issues and the competency form title is now a consistent heading. Wish I could test if it actually works though...so that's from the other backend ticket right?

webapp/src/components/CreateOrUpdateCompetencyForm.tsx
```diff
@@ -4,32 +4,47 @@ import React, { FormEventHandler, useState } from 'react';
...
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify({
-        label: competencyNameEntryText,
-        description: descriptionEntryText,
-      }),
-    })
+
+    fetch(
+      competencyId
+        ? `${process.env.REACT_APP_API_ORIGIN}/competencies/${competencyId}`
```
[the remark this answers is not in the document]
pthan1: No, when `competencyId` isn't set, the route was `/competencies/undefined`.  So, I had to add the ternary operation.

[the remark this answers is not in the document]
pthan1: Thanks for the solution.  I'll edit the code.

PR #306, opened 2022-02-11
Optionally get config from secrets files

### LINKED ISSUE #303: In order to use Docker Swarm secrets, api needs to read config from files
## Expected behaviour

The api service can be deployed to a Docker Swarm without any environment variables, and get its configuration from Docker secrets instead.

## Actual behaviour

The api service uses environment variables for external configuration and there is no option to get config from files.

## Details and resources

Docs on Docker secrets: https://docs.docker.com/engine/swarm/secrets/

## Checklist

davidvandusen:
## Proposed changes

Replaces uses of environment variables in the api with a service that first checks to see whether a Docker secret exists with the given key, then falls back to environment variables (for dev and test).

Resolves #303.

## Checklist

### DIFF
--- a/api/src/services/__tests__/configService.ts
+++ b/api/src/services/__tests__/configService.ts
@@ -1,0 +1,57 @@
+import { existsSync, readFileSync } from 'fs';
+
+import { clearConfig, findConfig } from '../configService';
+
+jest.mock('fs');
+const mockExistsSync = jest.mocked(existsSync);
+const mockReadFileSync = jest.mocked(readFileSync);
+
+describe('configService', () => {
+  describe('findConfig', () => {
+    it('should read a secrets file if one exists with the key name', () => {
+      const key = 'TEST_KEY';
+      const secretValue = 'secret value';
+      mockExistsSync.mockReturnValue(true);
+      mockReadFileSync.mockReturnValue(secretValue);
+      expect(findConfig(key, 'default value')).toEqual(secretValue);
+      expect(mockExistsSync).toHaveBeenCalledWith(`/run/secrets/${key}`);
+      expect(mockReadFileSync).toHaveBeenCalledWith(
+        `/run/secrets/${key}`,
+        'utf-8'
+      );
+      clearConfig(key);
+    });
+
+    it("should read the config from the environment if a secrets file doesn't exist with the file name", () => {
+      const key = 'TEST_KEY';
+      process.env[key] = 'environment value';
+      mockExistsSync.mockReturnValue(false);
+      expect(findConfig(key, 'default value')).toEqual(process.env[key]);
+      expect(mockExistsSync).toHaveBeenCalledWith(`/run/secrets/${key}`);
+      expect(mockReadFileSync).not.toHaveBeenCalled();
+      delete process.env[key];
+      clearConfig(key);
+    });
+
+    it('should use the default value if the config is nether a secret nor in the environment', () => {
+      const key = 'TEST_KEY';
+      const defaultValue = 'default value';
+      mockExistsSync.mockReturnValue(false);
+      expect(findConfig(key, defaultValue)).toEqual(defaultValue);
+      expect(mockExistsSync).toHaveBeenCalledWith(`/run/secrets/${key}`);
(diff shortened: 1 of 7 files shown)


BRashad (approved):

BRashad:
I approved the PR but maybe it is worth fixing the typo in the commit message. letter "L" is missed in the word "place".  Instead it is written "pace".

davidvandusen:
@BRashad Nice catch! I reworded the commit message.

eileenxue (approved):
Looks good to me, seems like you replace the `process.env.` to `findConfig`. What usually goes into the second parameter that's currently empty for some parts?

davidvandusen:
@eileenxue The second parameter is a default value, in case the key isn't in a secret or the environment. It's mainly for development environments. I decided to make it required to pass in the second parameter so that the default value (which in a lot of cases is just an empty string) is visible in the code where it's used.

api/src/services/__tests__/configService.ts
```diff
@@ -0,0 +1,57 @@
...
+      const secretValue = 'secret value';
+      mockExistsSync.mockReturnValue(true);
+      mockReadFileSync.mockReturnValue(secretValue);
+      expect(findConfig(key, 'default value')).toEqual(secretValue);
+      expect(mockExistsSync).toHaveBeenCalledWith(`/run/secrets/${key}`);
+      expect(mockReadFileSync).toHaveBeenCalledWith(
+        `/run/secrets/${key}`,
+        'utf-8'
+      );
+      clearConfig(key);
```
pthan1, line 22: Would an `afterEach` block be useful here to DRY up the test code?

api/src/services/__tests__/configService.ts
```diff
@@ -0,0 +1,57 @@
...
+      expect(findConfig(key, 'default value')).toEqual(process.env[key]);
+      expect(mockExistsSync).toHaveBeenCalledWith(`/run/secrets/${key}`);
+      expect(mockReadFileSync).not.toHaveBeenCalled();
+      delete process.env[key];
+      clearConfig(key);
+    });
+
+    it('should use the default value if the config is nether a secret nor in the environment', () => {
+      const key = 'TEST_KEY';
+      const defaultValue = 'default value';
```
pthan1, line 38: The variables here should be alphabetized.

api/src/services/__tests__/configService.ts
```diff
@@ -0,0 +1,57 @@
...
+      mockExistsSync.mockReturnValue(false);
```
pthan1, line 39: The test implies that the config isn't in the environment, but should it be more explicit (i.e. would it be valuable to set an environment variable at the `key` to a `null` value or `undefined`)?

api/src/services/__tests__/configService.ts
davidvandusen: In test code, readability and atomicity are frequently preferred over reducing duplication. `afterEach` is ideal for situations where a testing utility needs to hook some cleanup code after a test has run. Occasionally a group of examples may need similar setup and teardown, and if that group of examples are thought of as a unit (that is, they would change or be removed together) it may be appropriate to reduce duplication if there are many identical lines of code.

In this circumstance, the test writer is trying to make it clear to the reader that the memoized value needs to be cleared for each example to be considered complete, which would be obfuscated by the use of `afterEach`. (And, there is a whole other discussion about why the module memoizes itself instead of there being an intermediate layer.)

api/src/services/__tests__/configService.ts
davidvandusen: What value would be gained by swapping the order of these two declarations? Both are declared immediately before their first use, and the `key` variable is in the same position as in the other examples, which might aid readability.

api/src/services/__tests__/configService.ts
davidvandusen: The `process.env` global will never have a property set to `null` or `undefined` in normal circumstances. Either it will have an "own" property set to a string, or it won't. In the case where this test fails because the environment indeed has a non-empty string value for the test key, then it would be an exceptional case and this test would fail. That would be a good thing because it would reveal the exception instead of silently passing.

PR #309, opened 2022-02-15
Swap prod secrets from environment variables to Docker secrets

### LINKED ISSUE #308: The api still requires secrets to be passed via env vars in prod
## Expected behaviour

The api service gets secrets from Docker secrets in production instead of from environment variables.

## Actual behaviour

In order to deploy the api service, all the secrets need to be loaded into environment variables, which is an unnecessary additional step in deployment, and less secure than using Docker secrets.

## Checklist

davidvandusen:
## Proposed changes

Takes the environment variables that the production api service was using that contain secret values and configures them to be read from Docker secrets instead. 

Resolves #308.

## Checklist

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -1,20 +1,37 @@
 version: "3.7"
 
+secrets:
+  GITHUB_CLIENT_ID:
+    external: true
+  GITHUB_CLIENT_SECRET:
+    external: true
+  MYSQL_DATABASE:
+    external: true
+  MYSQL_HOST:
+    external: true
+  MYSQL_PASSWORD:
+    external: true
+  MYSQL_USER:
+    external: true
+  SESSION_SECRET:
+    external: true
+
 services:
   api:
     environment:
       - API_HOST=0.0.0.0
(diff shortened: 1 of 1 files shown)


BRashad:
Is this the best practice to when Docker is being used in the project we change the environment variables and configure them to be read from Docker? 
I wonder why this is done at this stage of the project, not for example later or earlier? Any best practice on when is the best time to do such changes?

davidvandusen:
@BRashad This ticket is part of a larger project to implement continuous deployment. I think the conversation about best practices isn't relevant to discuss on this PR since all it does is build on many incremental steps that have already been done and precedes more incremental work that is coming up.

If you want to learn about Docker secrets to see get an idea of why they're being used, a good start is the [Docker secrets docs](https://docs.docker.com/engine/swarm/secrets/).

BRashad:
ok no more questions then.

BRashad (approved):

eileenxue (approved):
Looks good to me. So many secrets, better security. Would be interested in reviewing the Docker Secrets page you provided in an earlier comment. Thanks!

PR #317, opened 2022-02-22
Display meeting notes for each meeting in the side drawer

### LINKED ISSUE #170: A button next to each meeting in the drawer should load and display the notes for that meeting
## Expected behaviour

Each meeting in the meetings drawer in the webapp has a plus icon button next to it. When that button is clicked, the notes associated with that meeting are loaded and displayed below the meeting heading. Clicking the button again hides the notes.

In the event of an error while fetching meeting data, the user is notified. While the data is loading, it is indicated to the user that the request for meeting data has been initiated.

## Actual behaviour

Users do not have a way to view meeting notes in the app.

## Details and resources

See the design spec for details.

https://docs.google.com/document/d/156DwfW14nBM2I8yAEFWOmmx4-zqX3kjzdgyuEzN5V1s/edit#heading=h.w8tawvdb54d4

There is already a `Meeting` component that loads and displays meeting notes.

The `MeetingDateTime` component can be refactored to be stateful and manage whether the `Meeting` is shown or not.

## Checklist

pthan1:
## Proposed changes
- Displays meetings as expandable accordion components in the meetings drawer.  
    - When accordions are expanded, agenda items/actions items appear.
- Changes MeetingDateTime component name to encompass change.

Resolves #170

Screen Shot 2022-02-22 at 4 28 14 PM

## Checklist


webapp/src/components/MeetingAccordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { formatDate, formatTime } from '../helpers/dateTime';
+import { Meeting as MeetingType } from '../types/api';
+import Meeting from './Meeting';
+
+interface MeetingAccordionProps {
+  meeting: MeetingType;
+}
+
+const MeetingAccordion = ({ meeting }: MeetingAccordionProps) => {
+  const [isInDrawer, setisInDrawer] = React.useState(true);
```
BRashad, line 22: @pthan1 settisInDrawer is never used. If it's not needed maybe we should get rid of it.

webapp/src/components/Meeting.tsx
```diff
@@ -28,9 +29,11 @@ const Meeting = ({ meetingId }: MeetingProps) => {
   )?.id;
   return (
     <Grid>
-      <h1>{`Meeting on ${formatDate(meeting.starts_at)} at ${formatTime(
-        meeting.starts_at
-      )}`}</h1>
+      {!isInDrawer ? (
+        <h1>{`Meeting on ${formatDate(meeting.starts_at)} at ${formatTime(
```
davidvandusen, line 33: Perhaps this heading should be in the `MeetingPage` component, and not the `Meeting` component. Then this prop wouldn't be necessary.

webapp/src/components/MeetingAccordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+  Link as MuiLink,
+  ListItem,
+  ListItemText,
+} from '@mui/material';
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import { Link as ReactRouterLink } from 'react-router-dom';
+import React from 'react';
+
+import { formatDate, formatTime } from '../helpers/dateTime';
+import { Meeting as MeetingType } from '../types/api';
```
davidvandusen, line 14: I've seen other files that alias api types to `APIMeeting`, and things like that. Would it be valuable to use that convention consistently?

davidvandusen (commented):
I'm curious what you think of my idea about moving the heading into the page component.

webapp/src/components/MeetingAccordion.tsx
pthan1: It's not needed, I'll take it out.

webapp/src/components/Meeting.tsx
pthan1: Yeah, I think this would be a good idea.  The meetings page does not fetch the meting info.  We could have a state variable on the declared on the `MeetingPage` component whose update function would be called from the `Meeting` component.  Would that be a good solution?

webapp/src/components/MeetingAccordion.tsx
pthan1: Yes, that would be valuable.

webapp/src/components/Meeting.tsx
davidvandusen: Sounds like we might want to consider moving to a "container components" architecture, where state is managed separately from presentation. Let's talk about this with the team.

LouisDG6: you mean like redux?

pthan1: Per our discussion, the conditional render of the heading (based on *where* the Meeting component lives) has been removed (screenshot updated) to avoid React anti-patterns.

LouisDG6 (approved):

davidvandusen (approved):

PR #321, opened 2022-02-24
Add form that allows users to add meeting notes

### LINKED ISSUE #171: A form in the meetings drawer should allow users to add meeting notes
## Expected behaviour

When a meeting in the meetings drawer in the webapp is expanded, there is a form with a text field and a submit button. When text is added in the field and the submit button is pressed, the text is posted to the api and saved. The data sent to the api is in the following format. Note that the `sort_order` is set to be 1 plus the `sort_order` of the last item in the current user's agenda (the list of notes with the `participant_id` associated with the `principal_id` of the current user). The `agenda_owning_participant_id` is set to the `participant_id` associated with the `principal_id` of the current user.

```json
{
  "agenda_owning_participant_id": 1,
  "sort_order": 2
}
```

Loading states and error handling are done as per the spec.

## Actual behaviour

Users do not have a way to create notes for meetings in the app.

## Details and resources

Refer to the Agenda Item Flow in the spec for details.

https://docs.google.com/document/d/156DwfW14nBM2I8yAEFWOmmx4-zqX3kjzdgyuEzN5V1s/edit#heading=h.c12m3v8evblw

## Checklist

LouisDG6:
## Proposed changes

Form allows users to add meeting notes. 
Resolves #171 

## Checklist

- [ ] Do all commit messages start with the issue number?

Screen Shot 2022-03-02 at 4 16 48 PM

Screen Shot 2022-03-02 at 4 11 49 PM

### DIFF
--- a/webapp/src/components/CreateMeetingNoteForm.tsx
+++ b/webapp/src/components/CreateMeetingNoteForm.tsx
@@ -1,24 +1,21 @@
-import { Alert, Card, CardContent, Snackbar, TextField } from '@mui/material';
+import { Alert, Snackbar, TextField, Typography } from '@mui/material';
 import { LoadingButton } from '@mui/lab';
-import React, { FormEventHandler, useState, useContext } from 'react';
+import React, { FormEventHandler, useState } from 'react';
 
 import { EntityId } from '../types/api';
-import SessionContext from './SessionContext';
-
 interface CreateMeetingNoteFormProps {
   agendaOwningParticipantId: EntityId;
   meetingId: EntityId;
   onMeetingNoteChanged?: (id: EntityId) => void;
-  greatestSortOrderOfMeetingNotesList: number;
+  sortOrder: number | undefined;
 }
 
 const CreateMeetingNoteForm = ({
   agendaOwningParticipantId,
   meetingId,
   onMeetingNoteChanged,
-  greatestSortOrderOfMeetingNotesList,
+  sortOrder,
 }: CreateMeetingNoteFormProps) => {
-  const { principalId } = useContext(SessionContext);
   const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
   const [saveMeetingNoteError, setSaveMeetingNoteError] = useState(null);
   const [isSuccessMessageVisible, setIsSuccessMessageVisible] = useState(false);
@@ -33,7 +30,7 @@ const CreateMeetingNoteForm = ({
       body: JSON.stringify({
         agenda_owning_participant_id: agendaOwningParticipantId,
         note_text: meetingNoteText,
-        sort_order: greatestSortOrderOfMeetingNotesList + 1,
+        sort_order: sortOrder,
       }),
     })
       .then(res => res.json())
@@ -59,36 +56,35 @@ const CreateMeetingNoteForm = ({
   };
   return (
     <form onSubmit={onSubmit}>
-      Add an agenda Item
-          <TextField
-            fullWidth
-            multiline
-            required
-            sx={{ 
-              mt: 2,
-              mb: 2, 
-            }}
-            onChange={event => setMeetingNoteText(event.target.value)}
-            value={meetingNoteText}
-          />
-        {saveMeetingNoteError && (
-          <div>
-            <Alert
-              onClose={() => setSaveMeetingNoteError(null)}
-              severity="error"
-            >
-              Item was not added.
-            </Alert>
-          </div>
-        )}
-          <LoadingButton
-            fullWidth
-            type="submit"
-            variant="contained"
-            loading={isSavingMeetingNote}
-          >
-            Save Agenda Item
-          </LoadingButton>
+      <Typography sx={{ mt: 2, fontWeight: 'bold' }}>
+        Add an agenda Item
+      </Typography>
+      <TextField
+        fullWidth
+        multiline
+        required
+        sx={{
+          mt: 2,
+          mb: 2,
+        }}
+        onChange={event => setMeetingNoteText(event.target.value)}
+        value={meetingNoteText}
+      />
+      {saveMeetingNoteError && (
+        <div>
+          <Alert onClose={() => setSaveMeetingNoteError(null)} severity="error">
+            Item was not added.
+          </Alert>
+        </div>
+      )}
+      <LoadingButton
+        fullWidth
+        type="submit"
+        variant="contained"
+        loading={isSavingMeetingNote}
+      >
+        Save Agenda Item
+      </LoadingButton>
       {isSuccessMessageVisible && (
         <Snackbar
           open={true}
--- a/webapp/src/components/CreateMeetingNoteForm.tsx
+++ b/webapp/src/components/CreateMeetingNoteForm.tsx
@@ -1,7 +1,6 @@
-import { Alert, Snackbar, TextField, Stack } from '@mui/material';
+import { Alert, Snackbar, Stack, TextField, } from '@mui/material';
 import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
 import { LoadingButton } from '@mui/lab';
-import { Button } from '@mui/material';
 import React, { FormEventHandler, useState } from 'react';
 
 import { EntityId } from '../types/api';
--- a/webapp/src/components/CreateMeetingNoteForm.tsx
+++ b/webapp/src/components/CreateMeetingNoteForm.tsx
@@ -82,6 +82,7 @@ const CreateMeetingNoteForm = ({
           </div>
         )}
           <LoadingButton
+            fullWidth
             type="submit"
             variant="contained"
             loading={isSavingMeetingNote}
--- a/webapp/src/components/CreateMeetingNoteForm.tsx
+++ b/webapp/src/components/CreateMeetingNoteForm.tsx
@@ -1,0 +1,112 @@
+import { Alert, Card, CardContent, Snackbar, TextField } from '@mui/material';
+import { LoadingButton } from '@mui/lab';
+import React, { FormEventHandler, useState, useContext } from 'react';
+
+import { EntityId } from '../types/api';
+import SessionContext from './SessionContext';
+
+interface CreateMeetingNoteFormProps {
+  meetingId: EntityId;
+  onMeetingNoteChanged?: (id: EntityId) => void;
+  greatestSortOrderOfMeetingNotesList: number;
+}
+
+const CreateMeetingNoteForm = ({
+  meetingId,
+  onMeetingNoteChanged,
+  greatestSortOrderOfMeetingNotesList,
+}: CreateMeetingNoteFormProps) => {
+  const { principalId } = useContext(SessionContext);
+
+  const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
+  const [saveMeetingNoteError, setSaveMeetingNoteError] = useState(null);
+  const [isSuccessMessageVisible, setIsSuccessMessageVisible] = useState(false);
+  const [meetingNoteText, setMeetingNoteText] = useState('');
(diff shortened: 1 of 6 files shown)


EnriqueTMT (approved):
Nothing suspicious here. I approve this message 👍

davidvandusen:
Just looking at the screenshot:

- Why isn't there any margin between the meeting notes and the form?
- Why is the form in a card, even though the sidebar doesn't have very much space for the extra padding?
- Why doesn't the text input span the full width of its container?

webapp/src/components/Meeting.tsx
```diff
@@ -34,6 +39,9 @@ const Meeting = ({ meetingId }: MeetingProps) => {
   const currentParticipantId = meeting.participants.find(
     ({ principal_id }) => principal_id === principalId
   )?.id;
+  const [greatestSortOrderNumber] = meeting.meeting_notes
```
davidvandusen, line 42: The meeting notes are already sorted when they come from the API. Also, it's not supposed to be based on all the notes, only the ones from the agenda of the current user.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const { principalId } = useContext(SessionContext);
+
+  const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
+  const [saveMeetingNoteError, setSaveMeetingNoteError] = useState(null);
+  const [isSuccessMessageVisible, setIsSuccessMessageVisible] = useState(false);
+  const [meetingNoteText, setMeetingNoteText] = useState('');
+  const onSubmit: FormEventHandler = event => {
+    event.preventDefault();
+    setIsSavingMeetingNote(true);
+
```
davidvandusen, line 28: 👾

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+    fetch(`${process.env.REACT_APP_API_ORIGIN}/meetings/${meetingId}/notes`, {
+      method: 'POST',
+      credentials: 'include',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        agenda_owning_participant_id: principalId,
+        note_text: meetingNoteText,
+        sort_order: greatestSortOrderOfMeetingNotesList + 1,
```
davidvandusen, line 36: Why not compute this before passing it in? Then the prop can just be called `sortOrder`.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const onSubmit: FormEventHandler = event => {
+    event.preventDefault();
+    setIsSavingMeetingNote(true);
+
+    fetch(`${process.env.REACT_APP_API_ORIGIN}/meetings/${meetingId}/notes`, {
+      method: 'POST',
+      credentials: 'include',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        agenda_owning_participant_id: principalId,
```
davidvandusen, line 34: Why is a principal ID being set as a participant ID? Those are different kinds of objects, and they have different ID sequences.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+        note_text: meetingNoteText,
+        sort_order: greatestSortOrderOfMeetingNotesList + 1,
+      }),
+    })
+      .then(res => res.json())
+      .then(({ data, error }) => {
+        setIsSavingMeetingNote(false);
+        if (error) {
+          setSaveMeetingNoteError(error);
+        }
```
davidvandusen, line 44: Should there be an `else` block that hides the error if it saves successfully?

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <form onSubmit={onSubmit}>
+      <Card>
+        <CardContent sx={{ pb: 0 }}>
+          <h3>Add Meeting Note</h3>
+        </CardContent>
+        <CardContent sx={{ pt: 0 }}>
+          <TextField
+            required
+            sx={{ mb: 2, width: '50%' }}
```
davidvandusen, line 67: Why is this being set to half width?

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+        setSaveMeetingNoteError(error);
+      });
+  };
+  return (
+    <form onSubmit={onSubmit}>
+      <Card>
+        <CardContent sx={{ pb: 0 }}>
+          <h3>Add Meeting Note</h3>
+        </CardContent>
+        <CardContent sx={{ pt: 0 }}>
```
davidvandusen, line 64: Removing padding like this makes it look like the components are being misused.

davidvandusen (changes requested):

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+      .catch(error => {
+        setIsSavingMeetingNote(false);
+        setSaveMeetingNoteError(error);
+      });
+  };
+  return (
+    <form onSubmit={onSubmit}>
+      <Card>
+        <CardContent sx={{ pb: 0 }}>
+          <h3>Add Meeting Note</h3>
```
davidvandusen, line 62: Which `<h2>` does this fall under? Should this even be a heading?

webapp/src/components/Meeting.tsx
davidvandusen: To clarify, this only needs to be a linear operation at worst. Sorting is not linear.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Not sure, it was taken from the `Comptency

webapp/src/components/CreateMeetingNoteForm.tsx
davidvandusen: So, if it should be a heading, why an `<h3>` nested under an unknown or absent `<h2>`? Headings ought to be nested correctly, and it appears as though this on assumes where the component will be mounted will result in correct nesting.

webapp/src/components/CreateMeetingNoteForm.tsx
eileenxue: Curious what does the emoji mean in this context? Is that alien emoji meaning for a bug? Or a foreign object?

davidvandusen: It's the "space invader" emoji. I'm just being silly. There's inconsistent white space here, so I thought I'd drop a subtle hint to manage white space more carefully.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: How do you mean?

davidvandusen: The components are designed to have padding to make them look correct. When used as intended, adjustments like this are unnecessary.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Would it be necessary to be in an `else` block or could it be outside?

davidvandusen: Outside of what? Do you mean "always hide the error, and then show it again if a new error occurs"?

Do whatever is correct in the situation. Right now, the component shows an error message even when the note successfully saves, which seems incorrect.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Thanks for clarifying!

webapp/src/components/CreateMeetingNoteForm.tsx
davidvandusen: (Also, would an improved label be enough? It seems strange to base this on `CreateOrUpdateCompetencyForm` since the two forms have dramatically different use cases.)

webapp/src/components/CreateMeetingNoteForm.tsx
davidvandusen: I don't understand this response. The system doesn't have "users". @pthan1: If by "user" you mean "principal", then that's the same as the current, incorrect code. I'm also unsure what you mean by "is the objet type to use" since both principal and participant ids are the same type: `EntityId`.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Yes, I was thinking of always hiding the error.  But now that I think about it, it would be friendlier to read if it was encapsulated in an `else` block.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Thanks for clarifying.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: I saw a Participant type that I thought could be used.  However, looking at the `Meeting` component, there is a `currentParticipantId` constant, should I pass that down as props to the form to use to set the participant ID for the POST request?

davidvandusen: Sounds reasonable.

webapp/src/components/Meeting.tsx
pthan1: Do you meant that the sort order should be based on the meeting notes with `authoring_partcipant_id`'s that match the current principal's ID?

davidvandusen: Participant ids and principal ids are not comparable. Issue #171 specifies:

> the `sort_order` is set to be 1 plus the `sort_order` of the last item in the current user's agenda (the list of notes with the `participant_id` associated with the `principal_id` of the current user).

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: > Just looking at the screenshot:
> 
> * Why isn't there any margin between the meeting notes and the form?
> * Why is the form in a card, even though the sidebar doesn't have very much space for the extra padding?
> * Why doesn't the text input span the full width of its container?
> * Why is there so much space between the text input and the button?

These issues have been addressed by reviewing the product spec and following those style guidelines.

pthan1:
> Just looking at the screenshot:
> 
> * Why isn't there any margin between the meeting notes and the form?
> * Why is the form in a card, even though the sidebar doesn't have very much space for the extra padding?
> * Why doesn't the text input span the full width of its container?
> * Why is there so much space between the text input and the button?

These issues have been addressed by reviewing the product spec and following those style guidelines.

webapp/src/components/Meeting.tsx
pthan1: Thanks!  The code was changed.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Thanks.  The new changes pass down the correct sort order to the form as opposed to computing the value in the form.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: The heading tags were removed to avoid any problems with incorrect nesting.

PR #326, opened 2022-03-03
Add sockets to meeting rooms

BRashad:
## Proposed changes

Add sockets to the meeting room  

## Checklist

### DIFF
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -5,7 +5,7 @@ import cors from 'cors';
 import { createClient as createRedisClient } from 'redis';
 import { createServer } from 'http';
 import express from 'express';
-import expressSession from 'express-session';
+import expressSession, { Session } from 'express-session';
 import helmet from 'helmet';
 import { Server } from 'socket.io';
 
@@ -27,6 +27,12 @@ import { findConfig } from './services/configService';
 declare module 'express-session' {
   interface Session {
     principalId: number;
+  }
+}
+
+declare module 'http' {
+  interface IncomingMessage {
+    session: Session;
   }
 }
 
@@ -56,7 +62,6 @@ const start = async () => {
 
   io.on('connection', async socket => {
     socket.on('meeting:join', meetingId => {
-      //@ts-ignore
       const { principalId } = socket.request.session;
       if (participantExists(meetingId, principalId)) {
         socket.join(`meeting:${meetingId}`);
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -60,10 +60,10 @@ const start = async () => {
     },
   });
 
-  io.on('connection', async socket => {
-    socket.on('meeting:join', meetingId => {
+  io.on('connection', socket => {
+    socket.on('meeting:join', async meetingId => {
       const { principalId } = socket.request.session;
-      if (participantExists(meetingId, principalId)) {
+      if (await participantExists(meetingId, principalId)) {
         socket.join(`meeting:${meetingId}`);
         socket.emit('meeting:joined', meetingId);
       }
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -18,6 +18,7 @@ import {
 } from './middleware/errorHandlingMiddleware';
 import { loggedIn } from './middleware/authMiddleware';
 import meetingsRouter from './middleware/meetingsRouter';
+import { participantExists } from './services/meetingsService';
 import questionnairesRouter from './middleware/questionnairesRouter';
 import reflectionsRouter from './middleware/reflectionsRouter';
 import settingsRouter from './middleware/settingsRouter';
@@ -52,6 +53,22 @@ const start = async () => {
       credentials: true,
     },
   });
+
+  io.on('connection', async socket => {
+    socket.on('meeting:join', meetingId => {
+      //@ts-ignore
+      const { principalId } = socket.request.session;
+      if (participantExists(meetingId, principalId)) {
+        socket.join(`meeting:${meetingId}`);
+        socket.emit('meeting:joined', meetingId);
+      }
+    });
+    socket.on('meeting:leave', meetingId => {
+      socket.leave(`meeting:${meetingId}`);
+      socket.emit('meeting:left', meetingId);
+    });
+  });
+
   const RedisStore = connectRedis(expressSession);
   const redisClient = createRedisClient({
     host: findConfig('REDIS_HOST', 'localhost'),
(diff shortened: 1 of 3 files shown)


api/src/server.ts
```diff
@@ -52,6 +59,21 @@ const start = async () => {
       credentials: true,
     },
   });
+
+  io.on('connection', async socket => {
+    socket.on('meeting:join', meetingId => {
+      const { principalId } = socket.request.session;
+      if (participantExists(meetingId, principalId)) {
```
davidvandusen, line 66: `participantExists` is an async function, so this only evaluates as a truthy value because it's a Promise object, which will always be truthy.

The function on line `64` must be marked `async`, and this call to `participantExists` must be marked `await`.

api/src/server.ts
```diff
@@ -52,6 +59,21 @@ const start = async () => {
       credentials: true,
     },
   });
+
+  io.on('connection', async socket => {
```
davidvandusen, line 63: This callback function doesn't have `await` in it, so `async` is not necessary.

api/src/services/meetingsService.ts
```diff
@@ -104,3 +104,14 @@ export const createMeetingNote = async (
   });
   return { id };
 };
+
+export const participantExists = async (
+  meetingId: number,
+  principalId: number
+) => {
+  const [participant] = await db('participants').select('principal_id').where({
```
davidvandusen, line 112: I don't understand why this selects `principal_id`. Is there a more intuitive thing this query could select?

api/src/services/__tests__/meetingsService.ts
```diff
@@ -110,6 +111,19 @@ describe('meetingsService', () => {
...
+  describe('participantExists', () => {
+    it('should query for the participant with the given id and associated meeting with given id', async () => {
+      const meetingId = 2;
+      const principalId = 3;
+      mockQuery(
+        'select `principal_id` from `participants` where `meeting_id` = ? and `principal_id` = ?',
+        [meetingId, principalId],
+        [3]
+      );
+      expect(await participantExists(meetingId, principalId)).toEqual(true);
```
davidvandusen, line 123: The test only checks for the happy path. Should there also be a test for the sad path?

davidvandusen (changes requested):
Looks like we missed a critical bug. Oops!

api/src/server.ts
BRashad: Ok will fix it now

api/src/server.ts
BRashad: ok will amend it

api/src/services/__tests__/meetingsService.ts
BRashad: ok will add the test now

api/src/services/meetingsService.ts
BRashad: meetingID for instance ?

davidvandusen: I recall we had set it to just select `1`. Did that not work? Why not just select the `id` column?

BRashad: i replaced principal_Id with meeting_Id

api/src/server.ts
```diff
@@ -52,6 +59,21 @@ const start = async () => {
...
+    socket.on('meeting:join', meetingId => {
+      const { principalId } = socket.request.session;
+      if (participantExists(meetingId, principalId)) {
+        socket.join(`meeting:${meetingId}`);
+        socket.emit('meeting:joined', meetingId);
+      }
+    });
+    socket.on('meeting:leave', meetingId => {
+      socket.leave(`meeting:${meetingId}`);
```
eileenxue, line 72: Can you help me understand why leaving the room doesn't require checking who the participant is?

api/src/services/meetingsService.ts
BRashad: > I recall we had set it to just select `1`. Did that not work? Why not just select the `id` column?

it did i just think querying meeting_id is more self-explanatory than just adding id or 1.

api/src/server.ts
BRashad: I personally do not think we need to check the participant 2nd time upon room leaving. The major point is to check for security purposes participant identity and relevance to the meeting upon joining the room.

api/src/services/meetingsService.ts
davidvandusen: Querying for `1` makes it obvious that it's not querying for data from the table. By querying for `meeting_id` but then not using it, it's confusing why the data is being fetched at all. Querying for `id` at least says, "Just trying to figure out if this row exists."

Separately, some databases have an `EXISTS` keyword for this purpose. In ORMs, exists queries are often translated to `SELECT 1 FROM...` because it works across databases.

Does Knex have an `EXISTS` feature? If not, it makes sense to follow known conventions and to not include code that doesn't do anything.

api/src/server.ts
davidvandusen: In the event that someone is removed as a participant from a meeting while they are connected, if it checked whether they were still a participant, then they would be forever stuck in a room for a meeting they aren't associated to.

Anyway, even if that case never happens, there's no value to doing the extra work of checking. If a client wants to leave a room, why not just let them?

api/src/services/meetingsService.ts
BRashad: Ok, now I understand.

api/src/services/__tests__/meetingsService.ts
```diff
@@ -110,6 +111,29 @@ describe('meetingsService', () => {
...
   });
 
+  describe('participantExists', () => {
+    it('should query for the participant with the given id and associated meeting with given id', async () => {
+      const meetingId = 2;
+      const principalId = 3;
+      mockQuery(
+        'select `id` from `participants` where `meeting_id` = ? and `principal_id` = ?',
+        [meetingId, principalId],
+        [2]
```
davidvandusen, line 121: This is an invalid response for this query. This query will respond with an array of objects with `id` properties. This should be updated to respond with a value that matches with what Knex would actually return. For example:

```suggestion
        [{ id: 1 }]
```

davidvandusen (approved):
I'd recommend taking my suggested edit before merging.

api/src/server.ts
eileenxue: Okay I see, thanks for the clarifications!

eileenxue (approved):
All seems to be in order.

api/src/services/__tests__/meetingsService.ts
BRashad: ok will do it now

PR #327, opened 2022-03-03
Create form that allows users to add meeting notes

### LINKED ISSUE #171: A form in the meetings drawer should allow users to add meeting notes

pthan1:
## Proposed changes
Creates a form that allows users to add meeting notes. 
Creates a separate meeting note component that can be displayed in the drawer and that holds the new create meeting note form.

Resolves #171 

## Checklist

Screen Shot 2022-03-03 at 1 25 12 PM

Screen Shot 2022-03-03 at 1 28 02 PM


webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+  agendaOwningParticipantId,
+  meetingId,
+  onMeetingNoteChanged,
+  sortOrder,
+}: CreateMeetingNoteFormProps) => {
+  const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
+  const [saveMeetingNoteError, setSaveMeetingNoteError] = useState(null);
+  const [isSuccessMessageVisible, setIsSuccessMessageVisible] = useState(false);
+  const [meetingNoteText, setMeetingNoteText] = useState('');
+  const [isSuccessIndicator, setIsSuccessIndicator] = useState(false);
```
davidvandusen, line 24: The name of this state is confusing to me. It looks like it means "is success being indicated to the user", which suggests it should be called `isSuccessIndicated`.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const updateSaveMeetingNoteButton = () => {
```
davidvandusen, line 26: This function is only ever called once that I can find. Does it need to be extracted, or could it just be inlined?

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
+import { LoadingButton } from '@mui/lab';
+import React, { FormEventHandler, useState } from 'react';
+
+import { EntityId } from '../types/api';
+interface CreateMeetingNoteFormProps {
+  agendaOwningParticipantId: EntityId;
+  meetingId: EntityId;
+  onMeetingNoteChanged?: (id: EntityId) => void;
+  sortOrder: number | undefined;
```
davidvandusen, line 11: Optional props are specified using `?`, not `| undefined`. However, it looks like `sortOrder` is required to be a number based on the validation in the api, so this shouldn't be optional.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+      })
+      .catch(error => {
+        setIsSavingMeetingNote(false);
+        setSaveMeetingNoteError(error);
+      });
+  };
+  return (
+    <form onSubmit={onSubmit}>
+      <Stack spacing={1}>
+        <b>Add an agenda item</b>
```
davidvandusen, line 71: It's unusual that this is an inline element with siblings that are block level elements. Should this be wrapped in a block level element to be consistent?

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+          variant="contained"
+          loading={isSavingMeetingNote}
+          color={isSuccessIndicator ? 'success' : 'primary'}
+          startIcon={isSuccessIndicator ? <CheckCircleOutlineIcon /> : ''}
+        >
+          {isSuccessIndicator ? 'Saved' : 'Save Agenda Item'}
+        </LoadingButton>
+      </Stack>
+      {isSuccessMessageVisible && (
+        <Snackbar
```
davidvandusen, line 96: Is this Snackbar the specified way of showing the success indication? It looks like there are two different success indicators. Do we need both?

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+  return (
+    <Stack spacing={1}>
+      {meeting.meeting_notes.map((meetingNote, index, meetingNotes) => (
+        <React.Fragment key={meetingNote.id}>
+          {(index === 0 ||
+            meetingNote.agenda_owning_participant_id !==
+              meetingNotes[index - 1].agenda_owning_participant_id) && (
+            <b>
```
davidvandusen, line 47: This is an inline element with a block level sibling. Should this be wrapped in a block level element?

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,73 @@
...
+              {meetingNote.agenda_owning_participant_id === null
+                ? 'Action items'
+                : meetingNote.agenda_owning_participant_id ===
+                  currentParticipantId
+                ? 'Your agenda items'
+                : 'Their agenda items'}
+            </b>
+          )}
+          <ul>
```
davidvandusen, line 56: Is each meeting note list item wrapped in a separate list? Shouldn't items be grouped into the same list? Do they even need to be list items?

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: I'll make that change.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: No, the green success button is.  We don't need both.  I'll remove this bit.

webapp/src/components/MeetingQuickView.tsx
pthan1: No, lists would create bullet points at the beginning of each new list item.  

Block-level elements would place each meeting note on it's own separate line.  Each note was changed to be wrapped inside `<div>` tags.  however I believe a `<Box>` MUI component would allow us to include some padding to indent the notes.  What do you think?

davidvandusen: There is a roadmap item to allow each note to be indented at a different level. The `<Box>` component would be a good fit for this. Sounds like a good choice.

webapp/src/components/MeetingQuickView.tsx
pthan1: Yes, I'll make the change.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Thanks.  I removed the `| undefined`.  `sortOrder` is initialized to `1` or set to another value depending on the number of notes in an agenda.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Yes, I'll change it.

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,75 @@
+import { Stack, Typography } from '@mui/material';
+import React, { useContext, useState } from 'react';
+
+import { EntityId, Meeting as APIMeeting } from '../types/api';
+import CreateMeetingNoteForm from './CreateMeetingNoteForm';
+import SessionContext from './SessionContext';
+import useApiData from '../helpers/useApiData';
+
+interface MeetingQuickViewProps {
+  meetingId?: EntityId;
```
pthan1, line 10: Passing down `meetingId` is optional here but I don't think it should be optional.

pthan1: I've removed it.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+      }),
+    })
+      .then(res => res.json())
+      .then(({ data, error }) => {
+        setIsSavingMeetingNote(false);
+        if (error) {
+          setSaveMeetingNoteError(error);
+        } else {
+          setSaveMeetingNoteError(null);
+        }
```
davidvandusen, line 45: Would the following do pretty much the same thing?

```suggestion
        setSaveMeetingNoteError(error);
```

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+      })
+      .catch(error => {
+        setIsSavingMeetingNote(false);
+        setSaveMeetingNoteError(error);
+      });
+  };
+  return (
+    <form onSubmit={onSubmit}>
+      <Stack spacing={1}>
+        <Typography variant="body1" sx={{ fontWeight: 'bold' }}>
```
davidvandusen, line 65: Isn't `body1` the default value for `variant`? If so, is there any point specifying it?

webapp/src/components/MeetingAccordion.tsx
```diff
@@ -3,15 +3,16 @@ import {
   AccordionDetails,
   AccordionSummary,
   Link as MuiLink,
-  ListItem,
-  ListItemText,
+  Typography,
 } from '@mui/material';
+
```
davidvandusen, line 8: Normally there's only one empty line in the imports section between the absolute and relative imports.

webapp/src/components/MeetingAccordion.tsx
```diff
@@ -21,26 +22,42 @@ interface MeetingAccordionProps {
...
-          <MuiLink
-            component={ReactRouterLink}
-            to={`/meetings/${meeting.id}`}
-            underline="none"
-            sx={{ color: 'text.primary' }}
-            onClick={e => {
-              e.stopPropagation();
+      <AccordionSummary
+        expandIcon={
+          <IconButton>
```
davidvandusen, line 27: I'm curious what this `IconButton` does. The examples in the docs don't have it.

https://mui.com/components/accordion/

webapp/src/components/MeetingAccordion.tsx
```diff
@@ -21,26 +22,42 @@ interface MeetingAccordionProps {
...
+        <MuiLink
+          component={ReactRouterLink}
+          to={`/meetings/${meeting.id}`}
+          underline="none"
+          sx={{ color: 'text.primary' }}
+          onClick={e => {
+            e.stopPropagation();
+          }}
+        >
+          <Typography variant="body1">
```
davidvandusen, line 46: Here's another use of the `body1` variant. I just checked the api, and it is indeed the default: https://mui.com/api/typography/

webapp/src/components/MeetingAccordion.tsx
```diff
@@ -21,26 +22,42 @@ interface MeetingAccordionProps {
...
+            Meeting on {formatDate(meeting.starts_at)}
+          </Typography>
+          <Typography
+            variant="body2"
+            sx={{
+              opacity: 0.87,
```
davidvandusen, line 52: Why is the opacity being set here? Shouldn't this be using a colour variant to change it to an appropriate shade of grey? I think `text.secondary` is the lighter text colour.

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return (
+    <Stack spacing={2}>
+      <Stack spacing={1}>
+        {meeting.meeting_notes.map((meetingNote, index, meetingNotes) => (
+          <React.Fragment key={meetingNote.id}>
+            {(index === 0 ||
+              meetingNote.agenda_owning_participant_id !==
+                meetingNotes[index - 1].agenda_owning_participant_id) && (
+              <Typography variant="body1" sx={{ fontWeight: 'bold' }}>
```
davidvandusen, line 48: Just marking this use of the default `Typography` variant.

webapp/src/components/MeetingAccordion.tsx
pthan1: The `IconButton` creates a grey outline when hovering over the `ExpandMoreIcon`.  I noticed the grey outline in the product spec.  Should it be removed?

webapp/src/components/MeetingAccordion.tsx
pthan1: Thanks.  I'll remove the `body1` variants where needed.

webapp/src/components/MeetingAccordion.tsx
pthan1: Thanks for this.  I'll change the color.  

I only found the `Text legibility` documentation from MaterialDesign (see the Dark text on light backgrounds section) and followed the guidelines there.  (https://material.io/design/color/text-legibility.html#text-backgrounds).

webapp/src/components/MeetingAccordion.tsx
pthan1: Removed.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Yes, `body1` is the default.  No point in specifying it.  I'll remove it.

webapp/src/components/MeetingAccordion.tsx
davidvandusen: The product spec uses an entirely different icon, and a different approach to revealing the detailed content. I recall we decided to use a different icon from the spec so that it matched more closely with user expectations. I believe that the example in the MUI docs matches with user expectations, so it would make sense to follow that example.

webapp/src/components/MeetingAccordion.tsx
davidvandusen: Yes, that's guidance for Material Design, which Material UI follows by defining a text palette that has reduced opacity black in two different shades which they call `text.primary` and `text.secondary`. The way to follow those guidelines when using Material UI is by using the appropriate colour from the palette—they've already done the work for us.

pthan1: OK, thank you for explaining.

webapp/src/components/CreateMeetingNoteForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (error) {
+          setSaveMeetingNoteError(error);
+        } else {
+          setSaveMeetingNoteError(null);
+        }
+        if (data) {
+          setMeetingNoteText('');
+          setisSuccessIndicated(true);
+          setTimeout(() => {
+            setisSuccessIndicated(false);
```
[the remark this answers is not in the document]
pthan1: As of right now, the component is never unmounted whether or not the `Accordion` or the `MeetingQuickView` are visible or not.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: I don't believe so... The suggestion always sets the `SaveMeetingNoteError`, which could be `undefined` when the `POST` is successful, or an object.  That said, I don't think lines 43-45 are necessary.

webapp/src/components/MeetingAccordion.tsx
pthan1: Yes, this also confuses the user as to what to click to expand/collapse the accordion.  I'll remove it.

webapp/src/components/CreateMeetingNoteForm.tsx
pthan1: Does it matter if `SaveMeetingNoteError` is set to `undefined` or `null`?

webapp/src/components/MeetingQuickView.tsx
pthan1: Thanks.

webapp/src/components/CreateMeetingNoteForm.tsx
davidvandusen: How is it being used? Is its use dependent on it being exactly `null` or would `undefined` be no different?

pthan1: It's being used in the `catch` statement of the fetch call and to conditionally render the error message.  I don't think it matters either way.  To make the code a little more concise, I will initialize the variable to `undefined` in order to have only two conditions for this variable.

webapp/src/components/MeetingQuickView.tsx
pthan1: Removed.

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const [changedMeetingNoteIds, setChangedMeetingNoteIds] = useState<
+    EntityId[]
+  >([]);
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
+  const [isSuccessIndicated, setIsSuccessIndicated] = useState(false);
+  const [meetingNoteText, setMeetingNoteText] = useState('');
+  const [saveMeetingNoteError, setSaveMeetingNoteError] = useState(null);
+
+  const { data: meetingWithNotes, error } = useApiData<APIMeeting>({
```
davidvandusen, line 37: It caught me off-guard that this component both takes a `Meeting` as a prop, but also fetches the meeting again from the API. I can think of a number of way where this could be made more intuitive.

The problem is that this use of `useApiData` begs the question, "Doesn't this component already have all the meeting data? Why is it fetching the meeting data again?"

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,182 @@
...
+}
+
+const MeetingQuickView = ({ meeting }: MeetingQuickViewProps) => {
+  const { principalId } = useContext(SessionContext);
+  const [changedMeetingNoteIds, setChangedMeetingNoteIds] = useState<
+    EntityId[]
+  >([]);
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [isSavingMeetingNote, setIsSavingMeetingNote] = useState(false);
+  const [isSuccessIndicated, setIsSuccessIndicated] = useState(false);
```
davidvandusen, line 33: What was successful? This variable name forces me to have to look at its usage to understand what it means.

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const { data: meetingWithNotes, error } = useApiData<APIMeeting>({
+    deps: [changedMeetingNoteIds],
+    path: `/meetings/${meeting.id}`,
+    sendCredentials: true,
+  });
+  if (error) {
+    return <div>There was an error loading the meeting.</div>;
+  }
+  if (!meetingWithNotes) {
+    return null;
```
davidvandusen, line 46: Should the summary still be shown even while it's loading the notes? The summary only needs data from the props in order to be rendered.

webapp/src/components/MeetingQuickView.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <MuiLink
+          component={ReactRouterLink}
+          to={`/meetings/${meeting.id}`}
+          underline="none"
+          sx={{ color: 'text.primary' }}
+          onClick={event => {
+            event.stopPropagation();
+          }}
+        >
+          <Typography>Meeting on {formatDate(meeting.starts_at)}</Typography>
```
davidvandusen, line 112: What does the `Typography` component do when no props are passed?

BRashad (approved):
Ran the suggested features in my local environment. All works as per the ticket requirements.

davidvandusen (approved):
This works. There are a couple little awkward things about the maintainability of the code that we can chat about before merging.

webapp/src/components/MeetingQuickView.tsx
pthan1: Adding a new agenda item.  I'll have to change the variable name since this refers to behavior of the create meeting note form.

webapp/src/components/MeetingQuickView.tsx
pthan1: The `meeting` prop does not have any meeting notes.  The implementation is incomplete here and should do the second fetch only when the accordion is expanded.  

Could fetch the all meetings with all of their individual notes and pass all of the data down fromm the Parent component in order to only do one fetch call.  That would require a change to the `MeetingsService`.

davidvandusen: We had had a discussion about using the `deps` and `shouldFetch` arguments to the `useApiData` hook to address this. It appears that that was not implemented. Has that approach been discounted as a solution?

pthan1: No, I'll work on that solution.

webapp/src/components/MeetingQuickView.tsx
pthan1: Yes, this line was removed as part of some changes made from one of your conversations about using the `deps` and `shouldFetch` arguments of the `useAPIData` hook.

webapp/src/components/MeetingQuickView.tsx
pthan1: The `Typography` actual wraps the text in a `<p>` tag and adds some styling to the text (to the default `body` MUI variant styles).  Otherwise, the string is just rendered as the content within `<a>` tags.

PR #328, opened 2022-03-03
Ignore types for socket.request for now; it breaks the build

davidvandusen:
## Proposed changes

`ts-node` runs the tests just fine, but `tsc` breaks with the type that was defined.

https://github.com/OpenTree-Education/rhizone-lms/runs/5413941756?check_suite_focus=true

So, `@ts-ignore` to unblock the pipeline until we figure out what happened.

## Checklist

### DIFF
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -5,7 +5,7 @@ import cors from 'cors';
 import { createClient as createRedisClient } from 'redis';
 import { createServer } from 'http';
 import express from 'express';
-import expressSession, { Session } from 'express-session';
+import expressSession from 'express-session';
 import helmet from 'helmet';
 import { Server } from 'socket.io';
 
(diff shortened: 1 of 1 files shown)


eileenxue (approved):

api/src/server.ts
```diff
@@ -62,6 +56,7 @@ const start = async () => {
 
   io.on('connection', socket => {
     socket.on('meeting:join', async meetingId => {
+      // @ts-ignore
```
eileenxue, line 59: Should there be more explanation in the code about why we're using @ts-ignore for this section?

daveytech (approved):

api/src/server.ts
davidvandusen: Added documentation.

PR #331, opened 2022-03-07
Make all migrations run on Flyway 6

### LINKED ISSUE #325: Downgrade of flyway to 6 is causing issues with initial database migrations with docker
## Expected behaviour

When you build the docker image with migration files (or rebuild with new migrations), it's expected to run smoothly and create the appropriate database schemas. 

## Actual behaviour

When the docker image tried to build, there was an issue starting with migration file 7 that stated that there was an error with the SQL syntax which was not compatible with the MySQL server version possibly associated with the new flyway downgrade.

Screen Shot 2022-03-02 at 3 38 05 PM

## Steps to reproduce

As long as the codebase is still using flyway version 6: 
1. If it's the first time building a docker image, the issue should show up when it's doing migrations. 
2. Otherwise, to test this issue, delete the existing database for rhizone development locally and rebuild the docker image.  

## Details and resources

Flyway was recently downgraded from version 8 to 6 in #323 to solve another issue. As a result of this, it's possible that Flyway made it use an older version of MySQL which causes an error with SQL syntax starting at migration file 7. 

For a temporary fix, I changed the flyway version back to 8 in `rhizone-lms/db/Dockerfile`, rebuild it to run the migrations and then changed the version back to 6. Subsequent builds will run smoothly until a new migration file is created that ALSO has SQL syntax differences. You may need to repair or delete migration file 7 from the flyway schema history if it previously failed to run.

## Checklist

davidvandusen:
## Proposed changes

In order to make it so that new developers can run all the migrations using the downgraded Flyway version, one of the migrations had to be updated. Changing a migration file updates the checksum of that file, which will cause validation of the schema to fail when Flyway runs in production.

A new `db/tasks` folder was created to manage one-off tasks to be run against the production database. A new task was created to update the checksum of the changed migration in production so that it will pass validation.

Resolves #325.

## Checklist

### DIFF
--- a/db/migration/V007__backfill_journal_entry_reflections.sql
+++ b/db/migration/V007__backfill_journal_entry_reflections.sql
@@ -1,30 +1,30 @@
-CREATE PROCEDURE createReflectionsAndAssignToJournalEntries()
-BEGIN
-    loopLabel: LOOP
-        SET @countUnassignedJournalEntries = (
-            SELECT COUNT(*)
-                FROM journal_entries
-                WHERE reflection_id IS NULL
-        );
-        IF (@countUnassignedJournalEntries = 0) THEN
-            LEAVE loopLabel;
-        END IF;
-        INSERT INTO reflections (principal_id)
-            SELECT principal_id
-            FROM journal_entries
-            WHERE reflection_id IS NULL
-            ORDER BY journal_entries.id
-            LIMIT 1;
-        UPDATE journal_entries
-            SET reflection_id = LAST_INSERT_ID()
-            WHERE reflection_id IS NULL
-            ORDER BY journal_entries.id
-            LIMIT 1;
-    END LOOP loopLabel;
-END;
-
-CALL createReflectionsAndAssignToJournalEntries();
-
-DROP PROCEDURE createReflectionsAndAssignToJournalEntries;
+-- CREATE PROCEDURE createReflectionsAndAssignToJournalEntries()
+-- BEGIN
+--     loopLabel: LOOP
+--         SET @countUnassignedJournalEntries = (
+--             SELECT COUNT(*)
+--                 FROM journal_entries
+--                 WHERE reflection_id IS NULL
+--         );
+--         IF (@countUnassignedJournalEntries = 0) THEN
+--             LEAVE loopLabel;
+--         END IF;
+--         INSERT INTO reflections (principal_id)
+--             SELECT principal_id
+--             FROM journal_entries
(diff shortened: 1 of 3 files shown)


db/migration/V007__backfill_journal_entry_reflections.sql
```diff
@@ -1,30 +1,30 @@
...
-            WHERE reflection_id IS NULL
-            ORDER BY journal_entries.id
-            LIMIT 1;
-    END LOOP loopLabel;
-END;
-
-CALL createReflectionsAndAssignToJournalEntries();
-
-DROP PROCEDURE createReflectionsAndAssignToJournalEntries;
+-- CREATE PROCEDURE createReflectionsAndAssignToJournalEntries()
```
pthan1, line 1: Is commenting out the SQL query the "update" to the migration that you talk about in this PR's description?  I wonder what happens if you comment the SQL out, and if we need to make a new migration for this SQL task.

pthan1 (approved):

db/migration/V007__backfill_journal_entry_reflections.sql
davidvandusen: @pthan1:

> Is commenting out the SQL query the "update" to the migration that you talk about in this PR's description?

Yes. Any chance to the content of a migration will result in a different checksum.

> Does this essentially remove the migration and not allow it to run?

The migration has already been run on production and will never run again against that database. The only time this SQL will be run again is in development for freshly created databases. The data migration is not necessary for a fresh database, so it doesn't matter whether it's run or not.

The reason the code is commented out and not deleted is because these migrations are kept around for posterity—to show what's been run against production. This is a weird edge case, and we're approaching the solution in a hacky way because we can't use baseline migrations because those are a paid feature.

> could we re-write these statements to work with Flyway 6?

Flyway 6 is just using an older MySQL driver. All we'd have to do to make it run properly is update the driver in the Flyway image. That said, this data migration never needs to run again, so there's no point in doing that.

pthan1: This may be simplifying the issue, but will this PR be reverted if/when the Flyway bug is fixed and Flyway is upgraded to 8?

davidvandusen: I can't imagine any reason why we would. A migration that's been run is dead. The only reason we keep it around is to setup new dev environments, and the version of the migration in this PR is perfectly fine for that.

PR #336, opened 2022-03-08
Upon successful meeting note creation, socket message is emitted to the associated room

### LINKED ISSUE #175: Sockets in a meeting room should receive a message when a new note is created
## Expected behaviour

When a new meeting note is created by the `POST /meetings/:id/notes` route handler, after the note is successfully saved to the database, a `meeting_note:created` message in the following format is emitted to the Socket.IO room associated with the meeting, such as `meeting:123`.

```json
{
  "id": 1,
  "agenda_owning_participant_id": 2,
  "authoring_participant_id": 3,
  "sort_order": 4,
  "created_at": "2022-02-22T22:22:22.222Z"
}
```

## Actual behaviour

No messages are emitted on the socket when new meeting notes are created.

## Details and resources

There is nothing listening for this message type yet. There aren't even any sockets in the meeting room. This work can be done in parallel to the tickets that add those features.

## Checklist

eileenxue:
## Proposed changes

- When a meeting note is successfully created, it will emit a socket message to the room associated with the meeting. 
- Add test cases in `meetingService` to test for new query data and null values

image
Resolves #175 

## Checklist

### DIFF
--- a/api/src/middleware/meetingsRouter.ts
+++ b/api/src/middleware/meetingsRouter.ts
@@ -105,6 +105,9 @@ meetingsRouter.post('/:id/notes', async (req, res, next) => {
     );
     return;
   }
+
+  req.io.to(`meeting:${meetingId}`).emit('meeting_note:created', meetingNote);
+
   res.status(201).json(itemEnvelope(meetingNote));
 });
 
--- a/api/src/middleware/routerTestUtils.ts
+++ b/api/src/middleware/routerTestUtils.ts
@@ -2,9 +2,14 @@ import { agent, SuperAgentTest } from 'supertest';
 import { createServer } from 'http';
 import express, { NextFunction, Request, Response, Router } from 'express';
 import expressSession from 'express-session';
-import { findConfig } from './../services/configService';
-import { Server } from 'socket.io';
+import { Server, Socket } from 'socket.io';
+import { SuperAgent, SuperAgentRequest } from 'superagent';
 
+declare module 'supertest' {
+  interface SuperTest<T extends SuperAgentRequest> extends SuperAgent<T> {
+    io: Socket;
+  }
+}
 const getPrincipalId = jest.fn();
 
 export const mockPrincipalId = (principalId: number) => {
@@ -17,20 +22,16 @@ export const createAppAgentForRouter = (router: Router): SuperAgentTest => {
   const app = express();
 
   const server = createServer(app);
-  const io = new Server(server, {
-    cors: {
-      origin: [findConfig('WEBAPP_ORIGIN', '')],
-      credentials: true,
-    },
-  });
+  const io = new Server(server);
 
-  const sessionMiddleware = expressSession({
-    resave: true,
-    saveUninitialized: true,
-    secret: 'secret',
-  });
   app.use(express.json());
-  app.use(sessionMiddleware);
+  app.use(
+    expressSession({
+      resave: true,
+      saveUninitialized: true,
+      secret: 'secret',
+    })
+  );
   app.use((req, res, next) => {
     req.io = io;
     next();
@@ -43,5 +44,5 @@ export const createAppAgentForRouter = (router: Router): SuperAgentTest => {
     next();
   });
   app.use(router);
-  return agent(app);
+  return agent(server);
 };
--- a/api/src/middleware/routerTestUtils.ts
+++ b/api/src/middleware/routerTestUtils.ts
@@ -1,6 +1,9 @@
-import expressSession from 'express-session';
-import express, { NextFunction, Request, Response, Router } from 'express';
 import { agent, SuperAgentTest } from 'supertest';
+import { createServer } from 'http';
+import express, { NextFunction, Request, Response, Router } from 'express';
+import expressSession from 'express-session';
+import { findConfig } from './../services/configService';
+import { Server } from 'socket.io';
 
 const getPrincipalId = jest.fn();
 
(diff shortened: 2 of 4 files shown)


api/src/middleware/routerTestUtils.ts
```diff
@@ -12,14 +15,26 @@ export const sessionDestroyMock = jest.fn();
 
 export const createAppAgentForRouter = (router: Router): SuperAgentTest => {
   const app = express();
+
+  const server = createServer(app);
+  const io = new Server(server, {
+    cors: {
```
davidvandusen, line 21: The test server doesn't need CORS configuration since all the requests are made by tests, not browsers.

api/src/middleware/routerTestUtils.ts
```diff
@@ -12,14 +15,26 @@ export const sessionDestroyMock = jest.fn();
...
+      origin: [findConfig('WEBAPP_ORIGIN', '')],
+      credentials: true,
+    },
+  });
+
+  const sessionMiddleware = expressSession({
```
davidvandusen, line 27: Why was `sessionMiddleware` extracted to a variable?

api/src/middleware/routerTestUtils.ts
```diff
@@ -12,14 +15,26 @@ export const sessionDestroyMock = jest.fn();
 
 export const createAppAgentForRouter = (router: Router): SuperAgentTest => {
   const app = express();
+
+  const server = createServer(app);
```
davidvandusen, line 19: A server is created from the `app`, but `agent` is being called with `app`, not `server` below. Does this mean that the app agent won't actually handle attempted socket connections?

Does this server need to be created? Would a mock of `io` be sufficient for testing instead?

api/src/services/__tests__/meetingsService.ts
```diff
@@ -152,6 +152,11 @@ describe('meetingsService', () => {
         [agendaOwningParticipantId, 6, noteText, sortOrder],
         [meetingNoteId]
       );
+      mockQuery(
+        'select `id`, `agenda_owning_participant_id`, `authoring_participant_id`, `sort_order`, `note_text`, `created_at` from `meeting_notes` where `id` = ?',
+        [meetingNoteId],
+        [{ id: 5 }]
```
davidvandusen, line 158: The mocked SQL and this response don't match. The query selects six fields, but only `id` is in the object in the response.

api/src/services/__tests__/meetingsService.ts
```diff
@@ -162,7 +167,106 @@ describe('meetingsService', () => {
         )
       ).toEqual({ id: meetingNoteId });
     });
-
+    it('should resolve to null if no meeting note id exists', async () => {
```
davidvandusen, line 170: I don't understand what situation this description is referring to. It looks like it's the situation where the insert statement doesn't insert a new row. Is there a better way to communicate that to the reader?

api/src/services/__tests__/meetingsService.ts
```diff
@@ -162,7 +167,106 @@ describe('meetingsService', () => {
...
+        await createMeetingNote(
+          meetingId,
+          principalId,
+          agendaOwningParticipantId,
+          noteText,
+          sortOrder
+        )
+      ).toEqual(null);
+    });
+    it('should return a meetingNote object from a given meeting note id', async () => {
```
davidvandusen, line 196: This test description doesn't seem to describe the test case. The test case calls `createMeetingNote` and asserts that a newly created meeting note is returned. There's already a test case for that on line [138](https://github.com/OpenTree-Education/rhizone-lms/pull/336/files#diff-685314a9340c322ec0afa4fcb0607baa7c735cf90e3892e67dbf72c54686eb80R138) above. Is this test redundant?

api/src/middleware/routerTestUtils.ts
eileenxue: Okay, I will remove the CORS configs in this file.

api/src/middleware/routerTestUtils.ts
eileenxue: Looking at the code again and comparing it with the server code, we didn't actually need to extract it to a new variable since `sessionMiddleware` is not being used anywhere else in the test.

api/src/middleware/routerTestUtils.ts
eileenxue: I fixed it so that the agent is calling the server. Yes, the server will need to be created in this situation so that it can be used to test socket connections.

davidvandusen: Sounds good. 👍

api/src/services/__tests__/meetingsService.ts
eileenxue: Okay, you're right. I added additional variables to represent the `meetingNote` object which is also the response of that query. Then I updated the expected response for the test to also equal `meetingNote`.

api/src/services/__tests__/meetingsService.ts
eileenxue: Yes this test case is redundant after fixing the previous test case. This has now been removed.

api/src/services/__tests__/meetingsService.ts
eileenxue: Yes, there's a better way to communicate that to the reader. When the meeting note responds with an empty array, it should resolve to null.

api/src/middleware/meetingsRouter.ts
```diff
@@ -105,6 +105,9 @@ meetingsRouter.post('/:id/notes', async (req, res, next) => {
     );
     return;
   }
+
+  req.io.to(`meeting:${meetingId}`).emit('meeting_note:created', meetingNote);
```
pthan1, line 109: Why is `req` used instead of `res` here?

pthan1 (approved):
From the screenshot, it looks like the code is working.  Are there any console errors?

davidvandusen (approved):

api/src/middleware/meetingsRouter.ts
eileenxue: We're using req here because it's trying to send/emit a message out to the specified room, it's not responding to a request.

pthan1: That makes sense.  `io` was also placed in the `req` object in the `meetingsService`.

eileenxue:
@pthan1 I don't see any console errors when I'm testing it.

PR #338, opened 2022-03-09
Renew TLS certificates on production automatically

### LINKED ISSUE #251: Certbot isn't on cron
## Expected behaviour

Certbot automatically renews certs every 60 days.

## Actual behaviour

Certbot has to be run manually. We get a notification via email when certs are about to expire, but still have to manually run `certbot renew`

## Checklist

davidvandusen:
## Proposed changes

Every Sunday at 08:00, this new workflow stops the nginx service on production, runs `certbot renew`, then redeploys the production stack, which starts nginx back up.

Resolves #251.

## Checklist

- [ ] Are all code changes sufficiently tested?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -48,4 +48,4 @@ jobs:
           END
       - name: Deploy services
         run: |
-          cat docker-compose.production.yml | ssh production 'docker stack deploy --with-registry-auth --compose-file - rhizone-lms'
+          cat docker-compose.production.yml | ssh ${{ github.event.inputs.environment }} 'docker stack deploy --with-registry-auth --compose-file - rhizone-lms'
--- a/.github/workflows/renew_tls_certificates.yml
+++ b/.github/workflows/renew_tls_certificates.yml
@@ -2,8 +2,8 @@ name: Renew TLS Certificates
 
 on:
   schedule:
-# Every Sunday at 08:00
-    - cron: '0 8 * * 0'
+# Every Monday at 17:00 UTC, 09:00 PST (-08:00)
+    - cron: '0 17 * * 1'
 
 jobs:
   run_certbot_renew:
--- a/.github/workflows/renew_tls_certificates.yml
+++ b/.github/workflows/renew_tls_certificates.yml
@@ -1,0 +1,33 @@
+name: Renew TLS Certificates
+
+on:
+  schedule:
+# Every Sunday at 08:00
+    - cron: '0 8 * * 0'
+
+jobs:
+  run_certbot_renew:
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -48,4 +48,4 @@ jobs:
           END
       - name: Deploy services
         run: |
-          cat docker-compose.production.yml | ssh production 'docker stack deploy --with-registry-auth --compose-file - rhizone-lms'
+          cat docker-compose.production.yml | ssh ${{ github.event.inputs.environment }} 'docker stack deploy --with-registry-auth --compose-file - rhizone-lms'
```
davidvandusen, line 51: This slipped through a recent PR. This value should not be hard coded.

.github/workflows/renew_tls_certificates.yml
```diff
@@ -0,0 +1,33 @@
+name: Renew TLS Certificates
+
+on:
+  schedule:
+# Every Sunday at 08:00
+    - cron: '0 8 * * 0'
```
davidvandusen, line 6: If this takes the site down, there will be a midnight alert. Better to do this during waking hours.

eileenxue (dismissed):
Why do the TLS certificates need to be renewed every week?

davidvandusen:
@eileenxue It only _attempts_ to renew them every week. We use Let's Encrypt, whose certificates are valid for 90 days, and they recommend renewing 30 days before they expire. By attempting to renew them every week, it gives us a few chances to ensure that the automatic renewal goes through before they outright expire.

.github/workflows/renew_tls_certificates.yml
```diff
@@ -0,0 +1,33 @@
...
+            User $SSH_USER
+            IdentityFile ~/.ssh/production.key
+          END
+          ssh-keyscan -H $SSH_HOST >> ~/.ssh/known_hosts
+      - name: Renew certificates and redeploy
+        run: |
+          ssh production docker service rm rhizone-lms_nginx
+          ssh production certbot renew
+          echo "$GITHUB_TOKEN" | ssh production docker login ghcr.io --username ${{ github.actor }} --password-stdin
+          cat docker-compose.production.yml | ssh production docker stack deploy --with-registry-auth --compose-file - rhizone-lms
```
davidvandusen, line 33: In the event that a new version of a service was built and tagged `latest` since the last deploy, the new version will get deployed. If that new version requires a database migration to have been run, it will put the app into an inconsistent state.

Is it an option to run the `deploy.yml` workflow after renewal completes?

BRashad (approved):
No comments from my side. Looks straightforward.

Issue #179, opened 2021-12-03
The meeting page should add newly added meeting notes in response to WebSocket messages


davidvandusen:
## Expected behaviour

On the meeting route/page in the webapp, after connecting to the WebSocket and having successfully joined the meeting room, when receiving a `meeting_note:created` message, the note received in the payload of the message is added into list of notes in the appropriate position.

## Actual behaviour

When a `meeting_note:created` message is received, the UI doesn't make any updates.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

eileenxue:
@davidvandusen I'm trying to clarify if the actual behaviour mentioned in this ticket exists in the application. I can't find any mentions of `meeting_note:added` in the source code.

davidvandusen:
@eileenxue It doesn't exist in the application. It will be added in the ticket #175, though it looks like it uses `created` instead of `added`, so that might be a typo. Those tickets can be done in any order. Feel free to to the other ticket first if you think it makes sense to.

eileenxue:
@davidvandusen Thanks David, let me take a look at that ticket as well. Should we use `created` or `added` then?

eileenxue:
@davidvandusen I decided to tend to #175 first, it seems like it'll make more sense for me. Plus it deals with the backend whereas this deal with the frontend and the meeting page is still under development in another PR. Trying to avoid conflicts.

I will use the term `created` as mentioned in the other ticket unless you think otherwise.

BRashad:
To clarify. In this part of the ticket, `the note received in the payload of the message is added into the list of notes in the appropriate position` all it requires is to get the content/payload of the message, related to the note and render it in UI?

davidvandusen:
@BRashad The rendering of the notes is already handled in the same component that manages the data, so nothing will need to be passed anywhere.

In this ticket, the note from the `meeting_note:created` event needs to be added into the already existing list of notes that were fetched using `useApiData`. That may require managing the api data as local state, since the `useApiData` hook does not expose a way to modify the state.

BRashad:
Thanks for the clarification.

PR #340, opened 2022-03-10
Add newly created meeting notes to page when server pushes them

davidvandusen:
## Proposed changes

It does what it says on the tin.

Addresses #179.

## Checklist

### DIFF
--- a/webapp/src/components/Meeting.tsx
+++ b/webapp/src/components/Meeting.tsx
@@ -1,7 +1,8 @@
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Stack } from '@mui/material';
 
-import { EntityId, Meeting as APIMeeting } from '../types/api';
+import { compareMeetingNotes } from '../helpers/meetingsHelper';
+import { EntityId, Meeting as APIMeeting, MeetingNote } from '../types/api';
 import { formatDate, formatTime } from '../helpers/dateTime';
 import SessionContext from './SessionContext';
 import useApiData from '../helpers/useApiData';
@@ -13,9 +14,17 @@ interface MeetingProps {
 
 const Meeting = ({ meetingId }: MeetingProps) => {
   const socket = useSocket();
+  const [meetingNotes, setMeetingNotes] = useState<MeetingNote[]>([]);
   useEffect(() => {
     socket.emit('meeting:join', meetingId);
+    const handleNewMeetingNote = (meetingNote: MeetingNote) => {
+      setMeetingNotes(meetingNotes =>
+        [...meetingNotes, meetingNote].sort(compareMeetingNotes)
+      );
+    };
+    socket.on('meeting_note:created', handleNewMeetingNote);
     return () => {
+      socket.off('meeting_note:created', handleNewMeetingNote);
       socket.emit('meeting:leave', meetingId);
     };
   }, [socket, meetingId]);
@@ -25,6 +34,9 @@ const Meeting = ({ meetingId }: MeetingProps) => {
     path: `/meetings/${meetingId}`,
     sendCredentials: true,
   });
+  useEffect(() => {
+    meeting && setMeetingNotes(meeting.meeting_notes);
+  }, [meeting]);
   if (error) {
     return <div>There was an error loading the meeting.</div>;
   }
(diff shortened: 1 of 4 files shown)


webapp/src/components/Meeting.tsx
```diff
@@ -25,6 +34,9 @@ const Meeting = ({ meetingId }: MeetingProps) => {
     path: `/meetings/${meetingId}`,
     sendCredentials: true,
   });
+  useEffect(() => {
+    meeting && setMeetingNotes(meeting.meeting_notes);
```
pthan1, line 38: Why do we have to set Meeting Notes when `meeting` updates?  Doesn't the server push newly created notes to the client when added?

webapp/src/helpers/__tests__/meetingsHelper.ts
```diff
@@ -0,0 +1,87 @@
+import { compareMeetingNotes } from '../meetingsHelper';
+import { MeetingNote } from '../../types/api';
+
+describe('meetingsHelper', () => {
+  describe('compareMeetingNotes', () => {
+    it('should sort non-null participants first', () => {
+      expect(
+        compareMeetingNotes(
+          { agenda_owning_participant_id: 1 } as MeetingNote,
```
pthan1, line 9: Should the arguments passed into the `compareMeetingNotes` have the same structure/properties as the MeetingNote object type?

webapp/src/helpers/__tests__/meetingsHelper.ts
```diff
@@ -0,0 +1,87 @@
...
+          { agenda_owning_participant_id: null } as MeetingNote
+        )
+      ).toBeLessThan(0);
+    });
+    it('should sort null participants last', () => {
```
pthan1, line 14: What are `null` participants?  Should this description be a little more descriptive to describe the column/property that would be `null`?

pthan1 (approved):

webapp/src/components/Meeting.tsx
```diff
@@ -13,9 +14,17 @@ interface MeetingProps {
 
 const Meeting = ({ meetingId }: MeetingProps) => {
   const socket = useSocket();
+  const [meetingNotes, setMeetingNotes] = useState<MeetingNote[]>([]);
```
eileenxue, line 17: Clarification on how the `<MeetingNote>` is being used here, what's the purpose of the first empty array bracket inside it?

webapp/src/types/api.d.ts
```diff
@@ -62,6 +62,7 @@ export interface Meeting extends Entity {
 }
 
 export interface MeetingNote extends Entity {
+  created_at: string;
```
eileenxue, line 65: Should these be sorted alphabetically or does that not apply in this context?

eileenxue (approved):

webapp/src/types/api.d.ts
davidvandusen: Yeah, that looks like it was missed in an earlier code change. We don't have a lint rule that enforces it, so it's not worth worrying about. If we add a lint rule about it, it'll get fixed when that rule gets enforced.

webapp/src/helpers/__tests__/meetingsHelper.ts
davidvandusen: I agree, this description doesn't represent the way we think about the system. A better description would be something like, "should sort notes in participants' agendas before action items" or something like that.

webapp/src/helpers/__tests__/meetingsHelper.ts
davidvandusen: I can think of more cons for filling out the objects than pros. For example, only including properties that matter for the specific example makes it clearer how the example works. It also causes the test to fail in the event that the implementation changes to consider those other properties before it takes the specified ones into account. I'd rather have false negatives than false positives, and I prefer less code over more code.

webapp/src/components/Meeting.tsx
davidvandusen: Any type with square brackets after it represents an array containing only members of that type. So, `MeetingNote[]` means "an array containing only `MeetingNote` objects". Does that answer your question?

eileenxue: Yes, that answers my question, I was also looking at this [article](https://www.carlrippon.com/typed-usestate-with-typescript/) that seems to talk about the same thing about having a state that is strongly typed and how to deal with it.

webapp/src/types/api.d.ts
eileenxue: Okay, that makes sense. Thanks for the explanation.

Issue #348, opened 2022-05-17
Profile needs links to other LMS features


glenyingchua:
## Expected behaviour

User can see buttons to journal, competencies and progress underneath profile summary. User will be redirected to appropriate feature when they click on corresponding button.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

MahoMori:
I think we all agreed to show progress on the profile page, so our team decided not to implement a button to progress.

sheacronin:
We have created the code for the button links, but are waiting on issue #346 to be completed before making a pull request because our code will need to be inserted into the Profile component that they build out.

Issue #349, opened 2022-05-17
Profile view should have progress trackers


glenyingchua:
## Expected behaviour

Users will see a progress view with two types of trackers. First, a circular rating to showcase competencies; second, a linear bar to showcase progression through a program. 

The circular rating will be dependent on the already existing competency table. The linear progress table will be hardcoded, but designed to fetch progress from a curriculum table.

Note: Out of scope - will depend on a curriculum or programs table with start and end dates.  

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

denizedaza:
> The linear progress table will be hardcoded, but designed to fetch progress from a curriculum table.

The curriculum table will be a new `Programs` table with the `start_date` and `end_date` fields added to the database

denizedaza:
Until the #355 and #356 database issues are resolved, we will be using placeholders for the areas we would be connecting to the database in the progress tracker.

denizedaza:
@seidior to track the linear progress with `start_date` and `end_date` and display the progress as a percentage, I was looking at a third party plugin (either dayjs: https://day.js.org/en/ or if there is something similar to it). Would this be something we can use? What is the best way to parse and calculate the dates? 

My thought process was: 
Ideally, we would use the plugin to subtract the `start_date` from the `end_date` (or vice versa, dates are tricky) as an integer variable (called `time_period` for example). 
Then, we would have another variable called `current_date` to grab the current date, subtract it from the `end_date` as another integer. 
**Then** calculate `current_date` over `time_period` to get the percentage.

denizedaza:
> Until the #355 and #356 database issues are resolved, we will be using placeholders for the areas we would be connecting to the database in the progress tracker. Working on front-end aspects only.

Will split the database aspects of the trackers into a new ticket. Closing for front-end aspects resolved

PR #354, opened 2022-05-19
#348 Add profile links to lms features

MahoMori:
## Proposed changes

This resolves issue #348.

Add links to Journal and Competency features to LMS on the profile page. The Competency button linked to Competency page and the journal button linked to home page.
We did not create a link to progress bar as it will be on the profile page after #349 is solved.

## Screenshot
Screenshot 2022-05-19 135448

## Checklist

### DIFF
--- a/webapp/src/components/Profile.tsx
+++ b/webapp/src/components/Profile.tsx
@@ -31,107 +31,102 @@ const user = {
 
 const Profile = () => {
   return (
-    <div>
-      <Container fixed>
+    <Container fixed>
+      <Grid
+        container
+        sx={{
+          justifyContent: 'center',
+          alignItems: 'center',
+          display: 'flex',
+        }}
+        spacing={2}
+      >
         <Grid
-          container
+          item
+          xs={12}
+          md={4}
           sx={{
             justifyContent: 'center',
             alignItems: 'center',
             display: 'flex',
           }}
-          spacing={2}
         >
-          <Grid
-            item
-            xs={12}
-            md={4}
+          <Avatar
             sx={{
-              justifyContent: 'center',
-              alignItems: 'center',
-              display: 'flex',
+              width: 150,
+              height: 150,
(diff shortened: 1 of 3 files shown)


webapp/src/components/Profile.tsx
```diff
@@ -27,95 +31,119 @@ const user = {
 };
 
 const Profile = () => {
+  let navigate = useNavigate();
```
achichikalova, line 34: Why do we use the UseNavigation hook here?
What if use button component with href attribute instead:
`<Button
            component="a"
            sx={{ mr: 1, height: '40px' }}
            href={'/competencies'}
  >Competencies</Button>`

denizedaza (approved):
Ran the branch on my local machine, it works as intended and matches the ticket issue. The team already made a note that the progress button is missing and will be implemented in a future ticket. Also noticed linting fixes within code. 
Noted that package.json & yarn.lock files edited to account for Windows users and doesn't pose an issue when running on my machine (Mac)

achichikalova (changes requested):
Could you consider my suggestion with the Button component instead of using useNavigation hook, please? Thank you. :)
Good job!

webapp/src/components/Profile.tsx
MahoMori: I checked if the way you suggested works and it does. We will discuss which way is better with others on Friday.

MahoMori: I changed useNavigate hook to buttons with href attribute

achichikalova (changes requested):
Why are we wrapping the container with the <div> element?
We already have the Container element, we don't need the <div> element here.
Also, to be more consistent with the code, I suggest deleting the <div> wrapping.
Thank you.

achichikalova (approved):
All issues were resolved. 
The code is consistent with the whole project and works well.

Issue #357, opened 2022-05-25
Flyway doesn't reliably run migrations, especially on slow hypervisors


seidior:
## Observed behaviour

Flyway tries `$FLYWAY_CONNECT_RETRIES` (currently set to 10) times to connect to the MySQL server, regardless if the MySQL server has finished loading and is in a listening state. For sufficiently fast computers, Flyway may connect within those 10 times. However, if your host running Docker takes too long to get MySQL to a listening state, Flyway just fails to run.

## Expected behaviour

Based on the structure of the Docker Compose file, we expect the Flyway container to start shortly after the MySQL container starts, then we expect it to reliably execute the database migrations.

However, because it's executing based on container *start* and not when MySQL is listening, it does not reliably execute the migrations. Ideally, what would be best is if Docker waited for the MySQL container to start listening before it executed the Flyway container.

## Steps to reproduce

1. Grab a computer that is sufficiently slow, or set up a virtual machine to run Docker, especially one that doesn't natively support cgroups (see here Windows or Linux).
2. Run `docker-compose up` to bring the services up and watch to see if Flyway runs the database migrations.
3. Observe that Flyway may give an error such as:
```
Message    : Communications link failure
The last packet sent successfully to the server was 0 milliseconds ago. The driver has not received any packets from the server.
```

## Details and resources

Let's implement health checks for at least the MySQL container (ideally, all containers should have health checks, but let's leave that out of scope for now), then launch the Flyway container after the MySQL container has passed its health check. Let's maybe give it 12 x 10 seconds (so, two minutes?) before it gives up trying to connect to MySQL.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

seidior:
PR merged, issue should be resolved.

PR #358, opened 2022-05-25
#357 Improve Flyway Reliability

seidior:
This adds a health check to the MySQL container, and it modifies the Flyway container to tell Docker to wait until the MySQL container is *listening* before starting the Flyway container.

## Proposed changes

I'm proposing to modify the `docker-compose.yml` file we all use to bring up our local development environments to delay launching the Flyway container until the MySQL container is considered 'healthy'; that is, when the MySQL container is listening for incoming connections. This should improve the reliability of the database migrations immensely, as long as the MySQL daemon can start listening within 120 seconds (two minutes) from container start and respond to the health check in a way that Docker understands.

If and only if that's the case will Docker execute the Flyway container to perform the database migrations. If it can't contact MySQL within 120 seconds, `docker-compose` will error out instead of continuing to bring up the development environment.

This resolves issue #357.

## Testing

I tested this in both my macOS virtual machine and my Windows 11 virtual machine to success, but I'd like others on the team in their different environments to check to see if it breaks anything on their end.

## Checklist

- [x] Are there screenshots for UI changes? (Does not apply.)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,11 +27,17 @@ services:
     platform: linux/x86_64
     ports:
       - "3306:3306"
+    healthcheck:
+      test: mysql ${MYSQL_DATABASE} --user=${MYSQL_USER} --password='${MYSQL_PASSWORD}' --silent --execute "SELECT 1;"
+      interval: 10s
+      timeout: 10s
+      retries: 12
   flyway:
     build: ./db/
     command: migrate
     depends_on:
-      - mysql
+      mysql:
+        condition: service_healthy
     environment:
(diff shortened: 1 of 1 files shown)


achichikalova (approved):
Amazing. Thank you for finding and fixing it. 🥰

tenzinlekphell (approved):
Thank you for adding this functionality.
I have wondered whether flyway timing out is due to slow computers or MySQL not loading up because I believe I had an issue with MySQL not loading due to another process occupying PORT 3306. Therefore, Flyway would stop loading and my setup would not start.

PR #359, opened 2022-05-27
#349 Add progress bar and rating tracker component

achichikalova:
## Proposed changes
Addresses #349.
Users can see a progress view with two types of trackers. First, a circular rating to showcase competencies; second, a linear bar to showcase progression through a program.

The circular and linear progress are hardcoded for now but designed to fetch progress from a program table.

### Desktop
image
### Tablet/Mobile
image

## Checklist

### DIFF
--- a/webapp/src/components/CompetencyRatings.tsx
+++ b/webapp/src/components/CompetencyRatings.tsx
@@ -12,7 +12,8 @@ const CompetencyRatings = () => {
    *
    * @privateRemarks
    * For the competency rating scale
-   *
+   * In the future, the rating data will be pulled from the database (competency table)
+   * Currently, this data is hardcoded
    */
 
   const rating = 3.6;
--- a/webapp/src/components/CompetencyRatings.tsx
+++ b/webapp/src/components/CompetencyRatings.tsx
@@ -1,0 +1,35 @@
+import { Rating, styled, Typography } from '@mui/material';
+import CircleIcon from '@mui/icons-material/Circle';
+import React from 'react';
+
+const CompetencyRatings = () => {
+  const StyledRating = styled(Rating)({
+    '& .MuiRating-iconFilled': {
+      color: '#1976d2',
+    },
+  });
+  /**
+   *
+   * @privateRemarks
+   * For the competency rating scale
+   *
+   */
+
+  const rating = 3.6;
+  return (
+    <>
+      <Typography component="h4" variant="h5" sx={{ mt: 5, mb: 2 }}>
+        Computational Thinking
+      </Typography>
+      <StyledRating
+        value={rating}
+        readOnly
+        icon={<CircleIcon />}
+        emptyIcon={<CircleIcon />}
+        precision={0.1}
+      />
+    </>
+  );
+};
+
+export default CompetencyRatings;
--- a/webapp/src/components/ProgressBar.tsx
+++ b/webapp/src/components/ProgressBar.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { LinearProgress, Tooltip, Typography } from '@mui/material';
+
+const ProgressBar = () => {
+  /** 
+
+   @privateRemarks
+   the dates are placeholders for now. The actual date values will be pulled from the database later on 
+   
+   months read as: 0 - 11 instead of 1 - 12
+
+   will later add a function to convert the date string into the Date format 
+   */
+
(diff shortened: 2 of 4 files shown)


webapp/src/components/__tests__/ProgressBar.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { render, screen } from '@testing-library/react';
+import ProgressBar from '../ProgressBar';
+import React from 'react';
+
+describe('testing if progress bar inside Profile component is rendering', () => {
+  it('should render progress bar', () => {
+    render(<ProgressBar />);
+    expect(screen.getByText('Learning Progress')).toBeInTheDocument();
+  });
```
sheacronin, line 9: Awesome job adding a test!

webapp/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+const ProgressBar = () => {
+  /**
+   * @privateRemarks
+   * the dates are placeholders for now. The actual date values will be pulled from the database later on
+   * months read as: 0 - 11 instead of 1 - 12
+   * will later add a function to convert the date string into the Date format
+   */
+
+  const start_date = new Date(2022, 4, 1).getTime();
+  const end_date = new Date(2022, 7, 30).getTime();
```
sheacronin, line 13: Are these values going to be brought in from the database later?

webapp/src/components/CompetencyRatings.tsx
```diff
@@ -0,0 +1,36 @@
+import { Rating, styled, Typography } from '@mui/material';
+import CircleIcon from '@mui/icons-material/Circle';
+import React from 'react';
+
+const CompetencyRatings = () => {
+  const StyledRating = styled(Rating)({
+    '& .MuiRating-iconFilled': {
```
sheacronin, line 7: What is this line doing? I'm just curious!

webapp/src/components/__tests__/ProgressBar.test.tsx
Nazanin-Rezaei, line 9: Do we need to catch errors here or not?

sheacronin (approved):
Thank you for your work on this! I think it's good that the Competency Rating and Progress Bar are separated out into their own, reusable components. I think the style is very clean and goes well with the app's current design.

Nazanin-Rezaei (approved):

MahoMori (approved):
I tested it on my local and it works perfectly. You prepared components for real data which solves the #349 issue.
I like how you thought about design by changing flex direction between row and column depending on the screen size.
It is amazing that you added test file!

webapp/src/components/ProgressBar.tsx
denizedaza: Ideally, yes, these are just placeholders for now. I also planned to add a function in the lines above that would parse however the `TIME` data is stored from the database into integers that the `Date` format could use. If I recall, in the database, it would be a stored in a `YYYY-MM-DD` TIMESTAMP string (?), so we would have the function parse this.

webapp/src/components/CompetencyRatings.tsx
denizedaza: If I understand the MUI docs correctly, it's customizing the way we use the `Rating` component from their library. In this case, it's letting us choose the colour fill to be blue instead of yellow, which I believe is the default.

webapp/src/components/__tests__/ProgressBar.test.tsx
tenzinlekphell: Thank you Shea! It was a great team effort for our very first test.
We did encounter an error during the push and fixed it.
I'm going to document the problem and solution and post it in google drive documentation folder.

tenzinlekphell: We wanted to test whether the ProgressBar component would render or not.
I don't think we need to catch errors here.
I hope I answered your question Nazanin.

achichikalova: Thank you, Shea!

webapp/src/components/CompetencyRatings.tsx
achichikalova: We are creating a styled component based on the Rating component here. It was the only way to change the default color (yellow) for the rating to blue. On the 7th line, we use the class name that Material UI created for the circular element.

Issue #350, opened 2022-05-17
Profile page should have a personalized greeting for the user


glenyingchua:
## Expected behaviour

The user will see a personalized greeting, depending on the time of day and user language, displayed above their name and profile picture.

Information can usually be grabbed from window.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

sheacronin:
We're looking into having custom greetings for the user based on their preferred language and were thinking about using a package for this. We found this as an example https://www.npmjs.com/package/translate, and were wondering if using something like this would be okay for the project?

seidior:
> We found this as an example https://www.npmjs.com/package/translate, and were wondering if using something like this would be okay for the project?

This seems to use Google Translate on the backend, and I have some thoughts about this:

- I think it might be slightly overkill to add GT for just a greeting.
- Using a machine translate process might be a bit fraught. Let's take English for example. "Hi", "Hello", "Hey there", "Greetings", "Salutations", "Good morning", and others are all roughly equivalent greetings, but where I'd say "hi" to everyone in the cohort, I would say "Good morning, Mr. President" to POTUS. So we would have to choose a base greeting that would give the same sense of formality and familiarity across different languages; however, with a machine translation service, it will probably default to a more formal translation such as 様 in Japanese, which would be the wrong vibe.

I think the better option would be to create an object that stores greetings, and we can accept PRs to extend the greetings to new languages as people see fit. This is a totally fine thing to hard-code. Something along the lines of:

```typescript
type GreetingType = {
    greeting: string;
    delimiter: string;
    beginning_punct: string;
    ending_punct: string;
    ltr: "ltr" | "rtl";
}

let morning_greetings = new Map<string, GreetingType>();
morning_greetings.set("en", {
  beginning_punct: "",
  greeting: "Good morning",
  delimiter: ", ",
  ending_punct: "!",
  ltr: "ltr"});
morning_greetings.set("es", {
  beginning_punct: "¡",
  greeting: "Buenos días",
  delimiter: ", ",
  ending_punct: "!",
  ltr: "ltr"});
morning_greetings.set("ar", {
  beginning_punct: "",
  greeting: "صباح الخير",
  delimiter: " ",
  ending_punct: "!",
  ltr: "rtl"});

const user_greeting = morning_greetings.get(current_user.preferred_language) ||
                      morning_greetings.get("en")!;

[...]

PR #360, opened 2022-05-27
#350 Personalized greetings for the user

MahoMori:
## Proposed changes

Closes issue #350.
Add personalized greetings for the user based on the time of the day and thier browser language. For code organization, we separated functions to the `greeting` file in the helper folder. We included English, French and Japanese for now. Anyone can add more in the future.
The greeting defaults to English if the browser language is not hardcoded.

## Screenshot
Screenshot 2022-05-27 135235

## Checklist

### DIFF
--- a/webapp/src/components/Profile.tsx
+++ b/webapp/src/components/Profile.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import {
   Container,
@@ -30,8 +30,36 @@ const user = {
 };
 
 const Profile = () => {
+  const getGreeting = () => {
+    const myDate = new Date();
+    const hrs = myDate.getHours();
+
+    let greet = '';
+
+    if (hrs < 12) greet = '🌅 Good Morning';
+    else if (hrs >= 12 && hrs <= 17) greet = '🌞 Good Afternoon';
+    else if (hrs >= 17 && hrs <= 24) greet = '🌇 Good Evening';
+
+    return greet;
+  };
+  const greeting = getGreeting();
+
   return (
     <Container fixed>
+      <Grid
+        container
+        sx={{
+          justifyContent: 'center',
+          alignItems: 'center',
+          display: 'flex',
+          mb: 4,
+        }}
+        spacing={2}
+      >
+        <Typography component="h2" variant="h6" color="primary">
+          Hello, {user.name}! {greeting}.
+        </Typography>
+      </Grid>
       <Grid
         container
         sx={{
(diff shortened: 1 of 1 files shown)


achichikalova (approved):
Fantastic work, guys! I love the way you implement it!

webapp/src/helpers/greeting.ts
```diff
@@ -0,0 +1,151 @@
...
+    user_greeting =
+      morning_greetings.get(preferredLanguage) || morning_greetings.get('en')!;
+  } else if (timeOfDay === 'afternoon') {
+    user_greeting =
+      afternoon_greetings.get(preferredLanguage) ||
+      afternoon_greetings.get('en')!;
+  } else if (timeOfDay === 'evening') {
+    user_greeting =
+      evening_greetings.get(preferredLanguage) || evening_greetings.get('en')!;
+  }
```
denizedaza, line 104: Would doing a `switch: case` accomplish something similar? Since you're checking the same variable for different results? Was just wondering if this is something you tried and decided against before settling on `if-else`

webapp/src/helpers/greeting.ts
```diff
@@ -0,0 +1,151 @@
+type GreetingType = {
+  greeting: string;
+  delimiter: string;
+  beginning_punct: string;
+  ending_punct: string;
+  ltr: 'ltr' | 'rtl';
+};
```
denizedaza, line 7: I do like this, great job!

webapp/src/helpers/greeting.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Gets hour based on Date() and returns the greeting string based on the different time of the day.
+ *
+ * @param username - user's name
+ * @returns The string which includes the icon and the greeting string depending on the time of the day (ex. "🌅 Good morning, {username}!")
+ *
+ */
+export const getGreeting = (username: string): string => {
+  const myDate = new Date();
+  const hrs = myDate.getHours();
```
denizedaza, line 125: If you're only using `myDate` once here, can you do `const hrs = new Date().getHours()` instead? Or are you planning to extract more data from `myDate` later?

denizedaza (approved):
Great job separating the logic into new files! Looks really good. I like the creativity and implementation of TypeScript features you used. I also ran it on my local machine and the feature does work as intended with no errors.

webapp/src/helpers/greeting.ts
MahoMori: I am sure it works! I am assuming this will be left to a future code organization.

webapp/src/helpers/greeting.ts
MahoMori: We don't use `myDate` anywhere else, so as you said, I can chain `.getHours()` to `Date()`.

Issue #370, opened 2022-06-21
Competencies table will need to be updated with a category_id


daveytech:
## Expected behaviour

Competencies table should be connected to categories through a category_id so competencies can be grouped and organized for reflections

## Details and resources

ERD to be included.

Migration 16.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

denizedaza:
Yesterday: 
- re-pulled from `main` after reverts were applied. Migration numbers were changed so updated branch to match
- deleted previous version of the branch (before reverts) and pushed updated version of branch to repo to look over
- related to #369 so will wait for approved PR for that issue before pushing this forward

Issue #369, opened 2022-06-21
Categories migration table should be added 


daveytech:
## Expected behaviour

The database schema will need to include a categories table that connects and groups competencies to be organized for different reflections.

The fields to include are:
- Label
- Description
- Image

## Details and resources

ERD
![Rhizone LMS_revised](https://user-images.githubusercontent.com/67711077/174881126-14b7b440-d683-4cf9-8dcd-5258ee2c26a7.png)

Migration 15.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #400, opened 2022-06-24
Update competencies table

denizedaza:
## Proposed changes

Closes issue #370 and is related to #369. This adds the foreign key of `categories_id` from the newly created `categories` table into the existing `competencies` table. This builds on and incorporates the latest ES Lint checks and the migration from #369

## Screenshots
Migration build on flyway:
image

Updated database:
image
image

## Checklist


daveytech (approved):
This satisfies the issue. Thank you for including the screenshots.

db/migration/V016__add_category_id_to_competencies.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE competencies
+  ADD COLUMN category_id BIGINT,
```
sheacronin, line 2: Do we want to add an `INDEX` here as well? I noticed that other tables that have foreign keys have done this (such as in the V_011 and V_006 migration files). I believe it is supposed to increase the speed of querying the table, but I'm not sure if it is necessary or not.

daveytech: Good Call @sheacronin ! @denizedaza Let's add it in to be consistent before merge.

denizedaza: On it!

denizedaza: If I'm understanding the `INDEX` property correctly, would it be something like this?
image

daveytech: Yes that fits the bill!

sheacronin (approved):
This code appears to add the `category_id` foreign key to the competencies table as intended, and the screenshots show the correctly updated database

Issue #428, opened 2022-11-19
The program activities should display on the frontend in a calendar representation


seidior:
## Description

Once the work for issue #427 is complete, we need to transform the graphical representation of the data in the frontend into a `react-big-calendar` display.

## Details and resources

The calendar should show all activities on the calendar in every view mode at appropriate times. Optionally, there should be a drop-down to select which program the user would like to view.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

camianderson:
Here's the current status of this ticket:

- The calendar is rendering on the calendar route in app.
- When we click in the event a pop up (window.alert) is showing up.
- In the pop up shows the title and the description of the activity.

Next step is to try to implement a modal to replace the window.alert.

seidior:
I'm breaking the dialog feature into another issue to simplify this ticket. Hopefully this makes life easier! Go ahead and remove the code that displays a JavaScript alert, clean up the code, and implement the changes from ticket #427 once its PR has been merged, then let's open a PR for this issue.

seidior:
Here's the latest status of the ticket:

- The PR is still in review, but changes were made to address the feedback from others.
- There are no outstanding to-do items for this ticket, but work will need to be done to address any additional issues that might pop up.
- It's up to date with `main`.

Issue #433, opened 2022-11-22
Clicking an event in the calendar should show a popup dialog with more information


seidior:
## Description

With the merger of issue #428, we will have a calendar display of all program activities for the user. However, the activities are not clickable to find out more information such as time range, description, etc.

The expectation is that an MUI Dialog will pop up showing the details of the activity. Clicking a close button or clicking in the shaded area outside of the Dialog will close the dialog. Clicking another activity will show the Dialog but with the details of that activity instead.

## Details and resources

You will need to use the Dialog component from MUI, and you will need to use React state management in order to handle the change of contents of the Dialog and whether or not it's open.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

seidior:
It's generally not advisable, because we want to keep each team's work isolated into each branch before merger in case we need to roll back or delete changes, but to keep things moving along, I've created a branch for this issue and incorporated the latest from #428 so that we can start work on this. But please be advised, any work on #428 may end up disrupting code you're working on here, so try to work on the parts of this (React state management and incorporating the MUI Dialog component, for example) that aren't tied strongly to the calendar, and leave the parts tying the dialog with the calendar until the end; hopefully PR #435 will be merged by then.

seidior:
Here's the latest status of this ticket:

- @ChainShoe-Z and @phoebehala got the MUI Dialog to display when clicking on an event. They came up with a great solution that works to show a dialog even if you change the program to a different program and then click an event.
- After all the major changes in #428, I had to adapt the code that was here to bring it up to date and preserve that working functionality as much as possible.

Here's what still needs to be done:

- I feel like the click handling could be done a little better instead of overloading one function with handling both open and close functionality.
- The dialog is only being passed the title of the event. It needs to be passed all the details we want to show in the dialog.
- The dialog needs much better styling, which will address the `// TODO` comment in `ActivityDialog.tsx`.

If you've previously worked on this ticket, you'll need to:
```shell
git pull
docker compose down
docker compose build
docker compose up
cd webapp
yarn develop
```
before working on this again because there were changes merged in from main that affect database migrations as well.

daveytech:
EOD Update:

Progress was made on rendering the data from event in the modal. Might have to be revised based on a possible redesign of the Program component. Further styling will be discussed and worked on.

[...]

Issue #427, opened 2022-11-19
The webapp should show a graphical representation of the programs and activities


seidior:
## Description

Once the work for issue #415 is complete, we should create a very basic view of the programs and activities returned from the API request as a stepping stone to having the calendar in place.

## Details and resources

This does not need to be fancy. It should just show the formatted details of the default program and its activities (with time in the browser's time zone, not UTC). This is a necessary prerequisite for showing the data in the calendar.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

seidior:
Update for today: team has completed work on this ticket, but that work cannot be merged until the work from ticket #415 is merged.

PR #435, opened 2022-11-22
#428 calendar view of activities

camianderson:
## Proposed changes

Display the calendar in the calendar route of the Frontend.
Inside of the calendar is displayed all the activities for the current program.

## Screenshots
Screen Shot 2022-11-22 at 4 16 56 PM
Screen Shot 2022-11-22 at 4 17 07 PM

## Next steps
- Create the logic for when the user click the activity in the calendar it pops up a modal with more information about the selected activity.
- Create an input to select what program calendar the user wants to see.

## Checklist


seidior:
Is there a concept of an all-day activity within `react-big-calendar`? How are we handling the different activities that are all-day events (like assignments) versus timed events (like class, standup, or retrospective)?

webapp/src/components/CalendarPage.tsx
```diff
@@ -1,9 +1,17 @@
...
+} from 'react-big-calendar';
 import { Container } from '@mui/material';
+import { DateTime, Settings } from 'luxon';
+import { ProgramWithActivities } from '../types/api';
 import React from 'react';
+import 'react-big-calendar/lib/css/react-big-calendar.css';
 import useApiData from '../helpers/useApiData';
 
-import { ProgramWithActivities } from '../types/api';
-import Program from './Program';
```
seidior, line 6: Also, I think that we're going to need to incorporate a <Program> component on the <CalendarPage> even if we're only showing one Program, we should at least see details about the program.

`Program.tsx` and `Activity.tsx` were added from #427, so you'll need to decide to either adapt the code in those files to work with the code you and Prabh wrote for #428 or to delete those files. There's no sense in keeping them around if they're not helpful for what we need to accomplish #428.

webapp/src/components/CalendarPage.tsx
```diff
@@ -1,9 +1,17 @@
+import {
+  Calendar,
+  luxonLocalizer,
+  Event as RBCEvent,
+} from 'react-big-calendar';
 import { Container } from '@mui/material';
+import { DateTime, Settings } from 'luxon';
+import { ProgramWithActivities } from '../types/api';
 import React from 'react';
+import 'react-big-calendar/lib/css/react-big-calendar.css';
```
seidior, line 10: Can we clean up some of these imports so they are sorted and separated between package imports and imports of other files within our webapp, similar to how other files have separated out these files?

webapp/src/components/CalendarPage.tsx
```diff
@@ -1,9 +1,17 @@
...
 import useApiData from '../helpers/useApiData';
 
-import { ProgramWithActivities } from '../types/api';
-import Program from './Program';
+Settings.defaultZone = 'America/Los_Angeles';
```
seidior, line 13: Perhaps we should set this to "America/Vancouver"?

webapp/src/components/CalendarPage.tsx
```diff
@@ -24,23 +32,25 @@ const CalendarPage = () => {
   if (!programs) {
     return null;
   }
+  const programEventsActivities: RBCEvent[] = programs[0].activities.map(
```
seidior, line 35: We should figure out a way that allows us to conform our activities to `react-big-calendar`'s `Event` type so that we can use them directly instead of having to transform them. Or if we can't, some way to make our front-end code work better with the native types within `react-big-calendar`.

wendyyng:
@seidior Optional in #428 to add a drop-down to select which program the user would like to view - is this something that we should expect from this PR? We only have one default program so I think showing a drop-down is not doing much for now but is beneficial if we are handling multiple programs.

webapp/src/components/CalendarPage.tsx
```diff
@@ -24,23 +32,25 @@ const CalendarPage = () => {
...
+    activity => {
+      return {
+        title: activity.title,
+        start: new Date(activity.start_time),
+        end: new Date(activity.end_time),
+        description: activity.description_text,
```
veranika-karpava, line 41: Do we need this line in Calendar? I think we didn't use it.

veranika-karpava (changes requested):
Great job! Everything works as expected in my local machine. 

Also, I think that activities that the duration equal to 0 and end_time: 24:00:00 do not display correctly in week, day and agenda. It doesn't related on this ticket and it'll be the next step
Week:
Screenshot 2022-11-22 at 16 34 23
Day: 
Screenshot 2022-11-22 at 16 35 50
Agenda: 
Screenshot 2022-11-22 at 16 36 46

phoebehala (commented):
Some issue about start time and end time might need to further discuss...
For events over than one day such as "Collaborative Work Period", "Independent Work Time", noticed it displayed '>>' in agenda view
image
image

For all-day events such as "Self-assessment", "Prototyping" (start time and end time set to the same), noticed it only displayed 12am .. wondering if the end time need to be changed to 23:59?
image
image

webapp/src/components/CalendarPage.tsx
camianderson: That is the information that is feeding the <Calendar>

webapp/src/components/CalendarPage.tsx
camianderson: Great idea!

webapp/src/components/CalendarPage.tsx
camianderson: I will do it!

seidior:
> @seidior Optional in #428 to add a drop-down to select which program the user would like to view - is this something that we should expect from this PR? We only have one default program so I think showing a drop-down is not doing much for now but is beneficial if we are handling multiple programs.

Not required for this ticket; I wouldn't have objected if it were here but it's not a problem that it isn't. I did mention in a comment on this PR that we should at least be showing some details of the program (like its name and maybe a start/end date), but I think the dropdown might be a good option for individual tickets, which we'll talk about more on Friday/Monday. Good question, @wendyyng!

webapp/src/components/CalendarPage.tsx
veranika-karpava: Thank you, Cami, the explanation!

wendyyng:
We found an error in the V20 migration. Line 99:  ('Independent Work Time', 'Work on the assigned ticket.', 6, 2, '11:00:00', '24:00:00', 780, 1, @curriculumId), should be week 6 day 4 instead of day 2.

wendyyng:
Further discussion needed for time zone settings
```
Settings.defaultZone = 'America/Vancouver';
const localizer = luxonLocalizer(DateTime);
```

wendyyng:
> Some issue about start time and end time might need to further discuss... For events over than one day such as "Collaborative Work Period", "Independent Work Time", noticed it displayed '>>' in agenda view image image
> 
> For all-day events such as "Self-assessment", "Prototyping" (start time and end time set to the same), noticed it only displayed 12am .. wondering if the end time need to be changed to 23:59? image image

'>>' in agenda view means until the end of day. So I think it's showing tht time correctly. It might be confusing to the users though.

For the all day activities, allDay prop is added to the event. If the duration is 0, we know it's an all day event.
```
const programEventsActivities: RBCEvent[] = activities.map(activity => {
    return {
      title: activity.title,
      start: new Date(activity.start_time),
      end: new Date(activity.end_time),
      description: activity.description_text,
      allDay: activity.duration === 0,
    };
  }
```
Documentation: http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--all-day-accessor

wendyyng:
The time zone settings has been removed for now since it is optional.
`Settings.defaultZone = 'America/Vancouver';`

wendyyng:
> Great job! Everything works as expected in my local machine.
> 
> Also, I think that activities that the duration equal to 0 and end_time: 24:00:00 do not display correctly in week, day and agenda. It doesn't related on this ticket and it'll be the next step Week: Screenshot 2022-11-22 at 16 34 23 Day: Screenshot 2022-11-22 at 16 35 50 Agenda: Screenshot 2022-11-22 at 16 36 46



> activities that the duration equal to 0 and end_time: 24:00:00 do not display correctly in week, day and agenda.

Settings.defaultZone = 'America/Vancouver';

> Great job! Everything works as expected in my local machine.
> 
> Also, I think that activities that the duration equal to 0 and end_time: 24:00:00 do not display correctly in week, day and agenda. It doesn't related on this ticket and it'll be the next step Week: Screenshot 2022-11-22 at 16 34 23 Day: Screenshot 2022-11-22 at 16 35 50 Agenda: Screenshot 2022-11-22 at 16 36 46

@veranika-karpava 
Activities that the duration equal to 0 and end_time: 24:00:00 do not display correctly in week, day and agenda:
- All day event with duration of 0 is showing correctly
- Events with end_time 24:00 are showing correctly in agenda view but not in week & day views. We would need to fix this in the migration instead of this PR - based on my discussion with Spencer

seidior:
program_activities_wednesday

I think I've addressed all the outstanding comments with my last contributed commits. Please pull the latest copy to your local, bring all your Docker containers down, rebuild all Docker containers, and bring them back up. Log in and check out the calendar page.

These latest commits should fix the following:

- Weird display of events that started at a certain time but went until the end of the day.

   These events were being displayed correctly, but they didn't make sense within the context of the program activities. The migration file was updated to address this issue and I force-merged those changes so as not to hold up this PR while I'm away.
   
- If there is more than one program, the program selector and the calendar will be displayed per program, causing multiple calendars to be displayed on the page.

   I've addressed this by moving around the code that was written earlier today by the team into different files so that nothing should get duplicated now. Go ahead, check it out! I actually implemented working functionality for the drop-down selector to update the calendar with different activities for different programs as @wendyyng suggested, and I added a second program in the migration file so that you'd be able to see this functionality work as expected.
   
- Some event durations do not match with (`end_time` - `start_time`)

   This is expected behavior; however, I made sure that my in my updates to the migration file, every duration matches the time between the start time and end time.

- The default time zone should be `America/Vancouver`.

   Previously this was being set in code in the frontend and also in the migration file. Wendy removed the line of code in the webapp, and the migration file was updated to use `America/Vancouver` instead.
   
- All day events were not showing as all day events.

   @wendyyng addressed this particular issue, and I've modified the code style slightly to the same end result.
   
- The week/day view didn't necessarily line up with the agenda view.

   With @wendyyng's fixes for allDay events and with my modifications to our migration file, all displays should appear the same. Of course, leave a comment and let us know if it doesn't.

I've also added some contributions of my own:

- In the process of working on #433, we found out the calendar has a way to scroll down automatically to a certain hour when looking at week or day view. The calendar now shows those views starting at 8:00AM (local time to your computer) instead of 12:00AM, but you can always scroll up inside the calendar to see earlier events if those were ever configured.
- I switched to week view by default, but we can switch it to anything we want. I'm open to good arguments for/against week view by default, and I've made it fairly easy to change this in the code if need be.
- I made the dropdown inline with the title (using MUI's version of a flexbox) to make the entire component look more cohesive. I'd love feedback on the change.

Final notes:

- The changes (and the base code before the changes) don't support mutability of the underlying program or its activities, so this will need to be addressed before adding any future "add activity" or "create program" features.
- There's more work to be done further enhancing the display of the calendar events and programs, but I think what we have now looks good.
- We may want to look into adding "time zone" as a user setting, but that change would need a database migration, changes to `api`, and changes to `webapp`.
- Thank you for the great discussion, @veranika-karpava, @phoebehala, and @wendyyng, and thanks for the leg work on making the calendar happen in the first place, @camianderson and @Prabhj33t. Thanks also to @SanazDn for catching up quickly, contributing, and testing during the day today and providing a fresh perspective. Unfortunately, that means more than half the cohort has worked on this ticket, so the number of people who can officially review this has dwindled in size. But regardless if you've worked on this issue or not, please do take a look at the functionality, the code changes, and the UI changes, and share any thoughts you might have, "official" approval or otherwise.

veranika-karpava (approved):
Amazing work! Everything works locally.
- checked yarn delint, it's ok
- checked yarn test, all passed
- checked web presentation in browser. Everything works as expected.
Also, I see that short title in week presentation doesn't wrap it to the next line. I think it should the separate issue.
Screenshot 2022-11-24 at 11 00 46

daveytech (approved):
Great solutions to incorporate the particularities for the package. I believe there are more opportunities to customize the Calendar but can be worked on through maintenance tickets. 

This PR satisfies issue #428 .

Issue #437, opened 2022-11-23
The sample data for the programs/activities feature contains errors that need to be adjusted


seidior:
## Expected behavior

The programs/activities feature got sample data in the V020 migration that is supposed to mimic the Professional Mentorship Program's activities.

## Actual behavior

While building out the calendar, we noticed a few errors where events were placed on the wrong day or at the wrong time, for instance.

## Details and resources

The migration file needs to be updated. I don't believe the migration has been run on the live site yet, so there shouldn't be any reason we can't update the migration file instead of creating a new one.

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #438, opened 2022-11-24
#437 Modifications to curriculum activities

seidior:
## Description

Resolves issue #437 not by only modifying the errors but by enhancing all the sample data with more thorough titles and descriptions. I wrote code to generate the SQL to insert for the example curriculum activities to ensure every activity is placed on the correct day and at the correct times.

Updates all the activities to better align their names, descriptions, and times with the current cohort’s activities, and add any missing activities.

Adds an additional curriculum value (“Custom Curriculum”) to allow for unstructured curricula.

Adds an additional program value (“Cohort 5”) with some non-final dates so we have more than one program with which we test the front-end features.

Updates the list of program activities to include more types of activities.

Improves spacing and layout so SQL is easier to debug if necessary.

Switches from `America/Los_Angeles` to `America/Vancouver` as the default time zone.

This was tested alongside simultaneous changes to the code in issue #428 and everything looks like it's in the right place.

## Checklist

### DIFF
--- a/db/migration/V019__create_programs.sql
+++ b/db/migration/V019__create_programs.sql
@@ -72,7 +72,12 @@ CREATE TABLE participant_activities (
 );
 
 INSERT INTO activity_types (title) VALUES
-  ('class'),
   ('assignment'),
+  ('all-day activity'),
+  ('planning'),
   ('standup'),
-  ('retrospective');
+  ('bulk development'),
+  ('office hours'),
+  ('workshop'),
+  ('retrospective'),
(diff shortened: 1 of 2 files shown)


seidior:
Force-merged this PR in order for another PR not to be held up. I would have liked some reviews on this before merging, but it got pretty thoroughly tested and because it's basically a content update, there's no real functionality to test, only copy edits. Given that it's basically a copy edit fix, any errors in this can be addressed in a similar PR later without holding anyone up.

Issue #445, opened 2022-11-28
Clicking '+ more' in the calendar's month view should show all the activities in the popup


wendyyng:
## Expected behaviour

Clicking '+ more' in the calendar's month view should show all the activities in the popup

## Actual behaviour

Clicking '+ more' redirects users to the day view.

## Details and resources

`react-big-calendar` provides an example for showing more via a popup: http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--example-6
<img width="510" alt="Capture" src="https://user-images.githubusercontent.com/71687298/204365946-36a03c97-6934-43fa-a850-3afbac7ea81c.PNG">

## Checklist

- [X] This issue is linked to the appropriate project.
- [X] This issue has appropriate labels.

PR #451, opened 2022-11-28
#445 Add popup in Calendar to enable the show more via popup feature in month view

wendyyng:
## Proposed changes

This PR resolves the #445 by enabling the `react-big-calendar` to show a popup with all the activities of the selected date when the user clicks '+ more' in the calendar's month view. The activities in the popup are also clickable and show the activity dialog with activity information when clicked.

## Checklist

image
image

### DIFF
--- a/webapp/src/components/ProgramCalendar.tsx
+++ b/webapp/src/components/ProgramCalendar.tsx
@@ -60,6 +60,7 @@ const ProgramCalendar = ({ program }: ProgramCalendarProps) => {
         getNow={() => DateTime.local().toJSDate()}
         scrollToTime={DateTime.local().set({ hour: 8, minute: 0 }).toJSDate()}
         style={{ height: 500 }}
+        popup
       />
       <ProgramActivityDialog
         show={dialogShow}


seidior:
I see this PR is still in draft status, but it's a simple enough change that I think it can be reviewed now.

Everything builds just fine on my machine. I merged in my latest changes from main (so you have one less task for you to do). Yarn didn't need to delint anything.

Two behaviors I observed that we can chat about in our meeting this morning:
- If I click the "+[x] more" link, then drag to resize the window to be more narrow, the popup doesn't move with the day in the calendar; it stays put.
- My expectation by clicking a "+[x] more" link would be that it would only show me the missing items, but it appears the default behavior is to show all items for the day.

If neither of these are addressable/configurable, that's totally fine, and I approve of this PR as it stands now.

wendyyng:
> 

**Popup doesn't move with the day in the calendar**
- You are right the popup stays put when the window is resized. The example on `react-big-calendar` also has the behaviour. It seems like there is no quick fix for this. I also took a look at Google Calendar, and their popups also behave this way.
d

- I also looked at the Issues and PR in `react-big-calendar` 's [GitHub repo](https://github.com/jquense/react-big-calendar). While there are issues related to popup, they are not exactly related to the issues we are having. Just in case if you are interested: 

1. [fix: update to current react-overlays](https://github.com/jquense/react-big-calendar/pull/2217) to constrain popup within the Month container
2. [Popup repositions to top-left of window on resize](https://github.com/jquense/react-big-calendar/issues/2018). The popup on the month view repositions to top left of window on resize. (Not the issue we have)

wendyyng:
> * "+[x] more" link would be that it would only show me the missing items

**"+[x] more" link would only show the missing items**
- Personally I thought it should include all the activities instead of just the missing items. But other people may have different opinion. 
- Again I looked at Google Calendar and their popup includes all activities.
e
c

seidior:
@wendyyng Awesome, thanks for doing that research! Then I agree, this is ready to go.

seidior (approved):

rrabinovitch (approved):
Changes appear as expected on my local machine, and I'm in agreement with Wendy re: having the popup include all activities!

Issue #453, opened 2022-11-29
Add the corresponding program for the activity pop out window


ChainShoe-Z:
## Expected behaviour

The activity pop out window should also display which program does this activity belong to.

## Actual behaviour

The pop out windows only show the activity type, date, and description.

## Steps to reproduce

_List steps that someone can perform to see the actual behavior, or delete this section if it's not relevant..._
After user clicking the activity, the program it belongs to should display next to the date

## Details and resources
![popout](https://user-images.githubusercontent.com/75060786/204885867-e7ab9919-8523-44c7-934e-ae416d5aa300.png)

Get the program from service function, then add the corresponding program next to the date.

      <DialogContent>
        <DialogContentText id="alert-dialog-description" sx={{ mb: 1 }}>
          {timeRange()}
        </DialogContentText>
        <DialogContentText id="alert-dialog-description">
          <b>Description:</b> {contents.description}
        </DialogContentText>
      </DialogContent>

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

PR #459, opened 2022-11-30
#453 Add program name in activity details Dialog

ChainShoe-Z:
## Proposed changes

_What does this pull request propose to change?_
Adding the corresponding program for the activity pop-out window
popout

## Checklist

### DIFF
--- a/webapp/src/components/ProgramActivityDialog.tsx
+++ b/webapp/src/components/ProgramActivityDialog.tsx
@@ -76,12 +76,14 @@ const ProgramActivityDialog = ({
           {timeRange()}
         </DialogContentText>
 
-        <DialogContentText><b>Program:</b>{contents.theProgram}</DialogContentText>
-
-        <DialogContentText id="alert-dialog-description">
-          <b>Description: </b> {contents.description}
+        <DialogContentText>
+          <strong>Program:</strong>
+          {contents.programTitle}
         </DialogContentText>
 
+        <DialogContentText id="alert-dialog-description">
+          <strong>Description:</strong> {contents.description}
+        </DialogContentText>
       </DialogContent>
 
       <Divider />
--- a/webapp/src/components/ProgramActivityDialog.tsx
+++ b/webapp/src/components/ProgramActivityDialog.tsx
@@ -19,11 +19,6 @@ interface ProgramActivityDialogProps {
   handleClose: () => void;
 }
 
-//todo
-// const findProgram = (programId: number) => {
-//   return
-// }
-
 const ProgramActivityDialog = ({
   show,
   contents,
--- a/webapp/src/components/ProgramActivityDialog.tsx
+++ b/webapp/src/components/ProgramActivityDialog.tsx
@@ -19,6 +19,11 @@ interface ProgramActivityDialogProps {
   handleClose: () => void;
 }
 
+//todo
+// const findProgram = (programId: number) => {
+//   return
+// }
+
 const ProgramActivityDialog = ({
   show,
   contents,
@@ -76,9 +81,12 @@ const ProgramActivityDialog = ({
           {timeRange()}
         </DialogContentText>
 
+        <DialogContentText><b>Program:</b>{contents.theProgram}</DialogContentText>
+
         <DialogContentText id="alert-dialog-description">
-          <b>Description:</b> {contents.description}
+          <b>Description: </b> {contents.description}
         </DialogContentText>
+
       </DialogContent>
 
       <Divider />
--- a/webapp/src/components/ProgramCalendar.tsx
+++ b/webapp/src/components/ProgramCalendar.tsx
@@ -30,7 +30,7 @@ const activitiesForCalendar = (
         end: new Date(activity.end_time),
         description: decodeHTML(activity.description_text),
         allDay: !activity.duration,
-        theProgram: correspondingProgramTitle//activity.program_id.toString()
+        programTitle: correspondingProgramTitle,
       } as CalendarEvent)
   );
 };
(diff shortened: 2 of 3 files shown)


webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -76,9 +76,12 @@ const ProgramActivityDialog = ({
           {timeRange()}
         </DialogContentText>
 
+        <DialogContentText><b>Program:</b>{contents.theProgram}</DialogContentText>
```
wendyyng, line 79: I think for consistency we should add a space after 'Program:'

webapp/src/types/api.d.ts
```diff
@@ -118,4 +118,6 @@ export interface CurriculumActivity extends Entity {
 
 interface CalendarEvent extends RBCEvent {
   description: string;
+  //todo
+  theProgram: string;
```
wendyyng, line 122: Personally I think 'program' without 'the' would suffice

wendyyng (changes requested):

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -76,9 +76,12 @@ const ProgramActivityDialog = ({
           {timeRange()}
         </DialogContentText>
```
camianderson, line 78: I think we can remove the extra spaces in lines 71, 73, 78, 80, 84, 86 and 88.

webapp/src/components/ProgramCalendar.tsx
```diff
@@ -27,6 +29,7 @@ const activitiesForCalendar = (
         end: new Date(activity.end_time),
         description: decodeHTML(activity.description_text),
         allDay: !activity.duration,
+        theProgram: correspondingProgramTitle//activity.program_id.toString()
```
seidior, line 32: Not only should comments like these be removed before a pull request is created, but I don't believe the comment is correct, since you want to save the program's name, not its ID.

Also, I would modify this member's name from `theProgram` to something like `programTitle`, or I would name it `program` and just attach the Program. But I think `programTitle` is the better approach to naming this member.

webapp/src/components/ProgramActivityDialog.tsx
seidior, line 79: I don't know how I didn't catch it before and that it got through PR, but in HTML we use `<strong></strong>` for emboldened emphasis; the use of `<b></b>` is either deprecated or discouraged.

If you end up taking Wendy's suggestion and adding a space, please make sure it's _after_ the `</strong>` and not before.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -76,9 +76,12 @@ const ProgramActivityDialog = ({
...
 
+        <DialogContentText><b>Program:</b>{contents.theProgram}</DialogContentText>
+
         <DialogContentText id="alert-dialog-description">
-          <b>Description:</b> {contents.description}
+          <b>Description: </b> {contents.description}
```
seidior, line 82: Did we need a space here before the closing tag? Browsers collapse multiple spaces between elements to only a single space, so I don't think this change to this line does anything, and I think it can be reverted.

And, as before, please modify this to use `<strong>` instead. I'm sorry I didn't catch that before.

webapp/src/types/api.d.ts
```diff
@@ -118,4 +118,6 @@ export interface CurriculumActivity extends Entity {
 
 interface CalendarEvent extends RBCEvent {
   description: string;
+  //todo
```
seidior, line 121: Only the most necessary of comments should be kept when the pull request is marked as ready for review.

webapp/src/components/ProgramActivityDialog.tsx
seidior: Agreed! Please always make sure you run `yarn delint` before pushing to GitHub :) thank you

seidior (commented):
Overall, you've created a solution that technically works, but it's a start and it's a work in progress. It seems like you've got a few things to address before this is ready to review again.

I'm going to mark my review as a "comment" review rather than requesting changes so that you don't have to wait for me when you address the opportunities to improve the code. However, I will be moving your pull request into draft status.

When you've addressed each of the points brought up by myself, Cami, and Wendy and you've run a `yarn delint` and committed and pushed those changes, then mark each of our notes as "mark resolved", and finally, take your PR out of draft status and let the rest of the channel know it's ready to be reviewed again.

Also, please be sure to keep an eye on changes to `origin/main` here on GitHub and keep your pull request up to date. You won't be able to merge your PR if it's behind `main` and not up to date.

Thank you for your work!

seidior:
@ChainShoe-Z I know you and Davey had some trouble amending the commit messages yesterday, and I had some time this morning, so I thought I would tackle amending your commits.

I rolled back to your "Remove comments and delint" commit and edited its message to include the issue number, then rebased it off the `origin/main` branch (rather than doing a merge commit, since I have to force push the changes anyway, so why not), then forced pushed the revised history to GitHub.

Fortunately, before I did the rebase off `origin/main`, I noticed that your commit also had removed our `.env.example` file, which would've resulted in your PR being rejected. So I had to actually edit your commit too, which is why it's unfortunately now attributed to me instead. Sorry about that; there doesn't seem to be a way around that.

Because I did a force push, if you want to make further edits to your branch on your development machine, you'll want to delete and reinstate your branch _before you make any changes_:

```shell
git checkout main
git branch -D 453-add-program-for-popout-window
git checkout 453-add-program-for-popout-window
```

If you don't perform those steps and you commit and force push any changes from your dev machine, you'll unfortunately restore the fixes I just put in :(

seidior (approved):
Requested changes have all been addressed by the author of the issue. I've rebased it again off `origin/main` to bring it up to date.

The functionality works as described, and I don't see any unintended behavior. I'm going to merge this in now since it doesn't need any additional review and so it doesn't further fall behind `origin/main`.

Thank you @camianderson and @wendyyng for reviewing the PR and leaving comments.

Issue #446, opened 2022-11-28
The calendar page needs a more explicit way to go back to the home page


camianderson:
## Expected behaviour

Another option for the User to click and be rerouted to the home page. 

## Actual behaviour

Right now the only option that the user has is to click in the title of the App.

## Steps to reproduce

- Create a design proposition to the issue
- Get approval with the Team
- Implement the design

## Details and resources

We can try to implement a back button using an HTML button, or anchor link or an image that will redirect to the main page.
Here are some resources about it:
- https://uxdesign.cc/design-with-the-web-in-mind-d9f9df2e8812
- https://designshack.net/articles/trends/website-button-design/
- https://www.smashingmagazine.com/2022/08/back-button-ux-design/

## Checklist

- [x] This issue is linked to the appropriate project.
- [x] This issue has appropriate labels.

camianderson:
I proposed to implement a "Back" home button in the top or in the bottom of the Calendar page as shown in the screen shots, but since this "Back" button behavior does not appears anywhere else in the application, we decided to don't add a new navigation element to the page.

![Screen Shot 2022-11-29 at 3 43 20 PM](https://user-images.githubusercontent.com/98124157/204647443-14c3caa4-b94a-4f8e-a086-97a57ce0d993.png)
![Screen Shot 2022-11-29 at 3 36 26 PM](https://user-images.githubusercontent.com/98124157/204647452-2a7a157e-3a73-46d0-b808-0974d3e61265.png)

camianderson:
11/29 progress update: I was able to Implement a new home icon in the Navbar so the user can go back home at any time.
I am working in figure out how to style the hovering in the title so we can indicate to the user that they can click in the title to go back to the home page.

camianderson:
11/30 progress update: I was able to add styling to the logo when it is hovered using Styled Components.

SanazDn:
Great job! I approved.

PR #460, opened 2022-11-30
#446 Add options for addressing "back"/"home" button feedback

camianderson:
## Proposed changes

Add a home icon to the Navbar for the User to click and be rerouted to the home page.
Add styling to the Rhizone logo so when the user hover the mouse over it so they will be more aware that they can use the logo to go to the main page.

https://user-images.githubusercontent.com/98124157/204907165-ea8d1670-3c36-40b4-9064-8a7ad8d3d07b.mov

## Checklist

### DIFF
--- a/webapp/src/components/Navbar.tsx
+++ b/webapp/src/components/Navbar.tsx
@@ -9,8 +9,10 @@ import PeopleIcon from '@mui/icons-material/People';
 import MeetingsDrawerContext from './MeetingsDrawerContext';
 
 const StyledLogo = styled.span`
-  &:hover {
+  cursor: auto;
+  h1:hover {
     color: #1565c0;
+    cursor: pointer;
   }
 `;
 
@@ -19,9 +21,9 @@ const Navbar = () => {
   return (
     <Grid alignItems="center" container px={3}>
       <Grid item xs>
-        <Link to="/" style={{ textDecoration: 'none', color: 'black' }}>
+        <Link to="/" style={{ textDecoration: 'none', color: 'black'}}>
           <StyledLogo>
-            <h1>Rhizone</h1>
+            <h1 style={{height: '50px', width:'125px'}}>Rhizone</h1>
           </StyledLogo>
         </Link>
       </Grid>
--- a/webapp/src/components/Navbar.tsx
+++ b/webapp/src/components/Navbar.tsx
@@ -8,7 +8,7 @@ import PeopleIcon from '@mui/icons-material/People';
 
 import MeetingsDrawerContext from './MeetingsDrawerContext';
 
-const StyledLogo = styled.ul`
+const StyledLogo = styled.span`
   &:hover {
     color: #1565c0;
   }
--- a/webapp/src/components/Navbar.tsx
+++ b/webapp/src/components/Navbar.tsx
@@ -21,9 +21,9 @@ const Navbar = () => {
   return (
     <Grid alignItems="center" container px={3}>
       <Grid item xs>
-        <Link to="/" style={{ textDecoration: 'none', color: 'black'}}>
+        <Link to="/" style={{ textDecoration: 'none', color: 'black' }}>
           <StyledLogo>
-            <h1 style={{height: '50px', width:'125px'}}>Rhizone</h1>
+            <h1 style={{ height: '50px', width: '125px' }}>Rhizone</h1>
           </StyledLogo>
         </Link>
       </Grid>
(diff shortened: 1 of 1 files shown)


webapp/src/components/Navbar.tsx
```diff
@@ -1,21 +1,34 @@
...
       <Grid item xs>
         <Link to="/" style={{ textDecoration: 'none', color: 'black' }}>
-          <h1>Rhizone</h1>
+          <StyledLogo>
+            <h1>Rhizone</h1>
+          </StyledLogo>
         </Link>
       </Grid>
       <Grid item xs="auto">
+        <IconButton sx={{ mr: 1 }} href="/">
```
wendyyng, line 29: I like the home button! And it works as expected :)

webapp/src/components/Navbar.tsx
```diff
@@ -1,21 +1,34 @@
...
 const Navbar = () => {
   const { open: openMeetingsDrawer } = useContext(MeetingsDrawerContext);
   return (
     <Grid alignItems="center" container px={3}>
       <Grid item xs>
         <Link to="/" style={{ textDecoration: 'none', color: 'black' }}>
-          <h1>Rhizone</h1>
+          <StyledLogo>
+            <h1>Rhizone</h1>
+          </StyledLogo>
```
wendyyng, line 25: I noticed that the location of the 'Rhizone' was moved a little to the right. I think it looks fine but just wondering if the rest of the team wants to change the location of the logo?

webapp/src/components/Navbar.tsx
```diff
@@ -1,21 +1,34 @@
...
 import PeopleIcon from '@mui/icons-material/People';
 
 import MeetingsDrawerContext from './MeetingsDrawerContext';
 
+const StyledLogo = styled.ul`
+  &:hover {
+    color: blue;
+  }
+`;
+
```
wendyyng, line 16: I like that it changes colour when you hover over the logo! Personally I think we might want to use the same blue colour of either the SIGN OUT text or the 'SAVE REFLECTION' button. I think it is the primary blue of the MUI. Also I know Spencer suggested using the @emotion/styled but I think MUI supports hover effects too: 
- https://mui.com/material-ui/customization/color/
- https://stackoverflow.com/questions/64983425/mui-button-hover-background-color-and-text-color

But otherwise everything works as expected and good job!

wendyyng (changes requested):

webapp/src/components/Navbar.tsx
camianderson: That is a great question! Let's talk about it in the next stand up. Thank you for noticing it!

webapp/src/components/Navbar.tsx
seidior: I agree with @wendyyng that the default blue is a bit harsher than the blue that we use in the rest of the app, but I will push back to say this blue does better match the links in the footer at the bottom of the page.

**If** we end up keeping this hover state to turn blue, I would ask Davey if he wants to make the links in the footer also match the "Sign Out" button and the rest of our app and use the... _friendlier_ blue. :)

That said, while we're talking about this, I'd like to point out that it's bad form to put `<ul>` (a block-level element) inside of an `<a>` (an inline-level element), and I don't understand why we'd want to use an unordered list here instead of the appropriate HTML element, `<span>`. With your code, this is what the resulting DOM element structure looks like:

```html
<a style="text-decoration: none; color: black;" href="/">
  <ul class="css-1km7vmx">
    <h1>Rhizone</h1>
  </ul>
</a>
```

But if you end up taking Wendy's suggestion to use Material UI's native components to handle this hover state, this becomes a moot point. Still, hopefully this is still valuable discussion nonetheless!

camianderson: Thank you @seidior and @wendyyng  for the feedback! Those are really good observations that I will research more, learn and implement if we agree to.

camianderson: As voted in the stand up meeting, the blue was changed to the same blue we use in the button in the main page.
I also changed the <ul> to a <span> as discussed above.

wendyyng (approved):
I approve this PR since everything works as expected on my local machine. Great work!

ChainShoe-Z:
Everything else works on my machine, but even I hover to space of the nav bar and click, the "Rhizone" button also works, I'm not sure if this is fine for this function. The home icon works perfect and very useful for users
1230e0b273d8dfd2556e19999f00111

camianderson:
Great catch @ChainShoe-Z! It is fixed! 😄

SanazDn:
Great job! I like it and approve it.

SanazDn (approved):

PR #486, opened 2022-12-13
#457 Mark Activities Complete/Incomplete in Calendar's ProgramActivityDialog

### LINKED ISSUE #457: The ability to mark a task as completed from the frontend needs to be developed
## Expected behavior

The last remaining big feature of the programs/activities component is the ability to mark activities (classified as "assignments") as completed and to have a visual indicator of which ones have been completed.

## Actual behavior

Right now in our Dialog component, it shows all of the details of the activity except its completion status, and there is no way to mark a task as completed.

## Details and resources

Minimum requirements for resolving this issue:

- When you click on a task and the Dialog element is presented, if the event is of type "assignment", the webapp should show the user's completion status (maybe with a checkbox?) in the Dialog.
- Clicking either the "mark completed" button or clicking a checkbox will send an API request to mark the task as completed.
   - If it's implemented as a button, there should be a different button that will now show to mark the task as "not completed", in case marking it completed was in error.
   - If it's implemented as a checkbox, unchecking the checkbox should send an API request to mark the task as not completed.

Stretch goals: (not required, but nice to have)

- There should be a visual indicator in the calendar for which activities are assignments and are either complete or not complete
- The button/checkbox to mark a task as complete should have a visible status to it to show the different states of the API request (like spinning or other animation), and the "completion" state should only be shown if the API request succeeds
- If we're unable to mark a task as completed, then there should be a visible error message displayed to the user so that they can act on it.

Reach goals: (not in scope, but if they are included, that'd be rad!)

- There should be a view that shows us a list of the activities that are assignments, their due dates, and their completion status. Think kind of like a to-do list view.

## Checklist

wendyyng:
## Proposed changes
This PR resolves #457.

Minimum Requirements:

This PR adds a 'Mark Complete/Incomplete' button to the ProgramActivityDialog when the activity type is an assignment. Users can mark activities as completed or incomplete by clicking the button. The frontend sends a GET request to the backend to retrieve the activity status when the ProgramActivityDialog is open and sends a PUT request with the updated status to the backend when the button is clicked.

Stretch:
The Task Alt Icon shows on the right of the activity title when the activity status is completed. This PR also adds the isLoading state for both GET and PUT requests, so the spinner shows in the button when the API request is loading. If the PUT request is successful, the success message will show in the toast/alert on the bottom left. Error handling is also included to show an error message in the toast/alert when errors occurred for both GET and PUT requests. When the GET request is not successful, the button will be disabled to avoid the user sending a PUT request.

Not required for this ticket but the close button on the bottom right is replaced by the small close icon on the top right.

## Checklist

image
image
image
image


Prabhj33t (commented):
The x button overlaps with the check sign here otherwise, all seems good. I approve of this ticket

Screenshot 2022-12-13 at 3 17 30 PM

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
```
seidior, line 1: Since we're importing `useState()` and `useEffect()` here, let's do the same in `ProgramCalendar.tsx` and replace all the `React.(function name)` calls in that file with just the function names, so that things appear the same in both files.

seidior: (We probably could also do the same with `Dispatch` and `SetStateAction`, too.)

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+  FormGroup,
+  Snackbar,
   Table,
   TableBody,
   TableCell,
   TableContainer,
   TableRow,
 } from '@mui/material';
-
+import IconButton from '@mui/material/IconButton';
```
seidior, line 23: Can we group this together with the other imports from `@mui/material`? Also, maybe a space after the `} from '@mui/material';` line and before the import from `'../helpers/...'` and then again after the import of `CalendarEvent`.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
 import { formatDate, formatTime } from '../helpers/dateTime';
 import { CalendarEvent } from '../types/api';
-
 interface ProgramActivityDialogProps {
   show: boolean;
   contents: CalendarEvent;
   handleClose: () => void;
 }
+const sendAPIGetRequest = (
```
seidior, line 31: Space above this ideally, please.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+) => {
+  setIsLoading(true);
+  if (activityType !== 'assignment') return;
+  return fetch(`${process.env.REACT_APP_API_ORIGIN}${path}`, {
+    method: 'GET',
+    credentials: 'include',
+  })
+    .then(res => res.json())
+    .then(
+      async ({ data }) => {
```
seidior, line 47: Does this need to be an async function? It isn't for the PUT request version of this function.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+        if (data) {
+          setIsLoading(false);
+          setError(false);
+          setCompleted(data.completed);
+        }
+      },
+      error => {
+        setIsLoading(false);
```
seidior, line 55: In `useApiData()`, we handle ignoring an AbortError, so maybe we should have that exception here. Also, it'd be great if we could pass the error message back to the toast so we can be more specific about the error we encountered, but it's not required within the scope of this ticket.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+        setError(true);
+        setIsErrorShown(true);
+      }
+    );
+};
+
+const sendAPIPutRequest = (
+  path: string,
+  body: Object,
```
seidior, line 64: Let's be more specific about the shape of this `Object` here.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
   contents,
   handleClose,
 }: ProgramActivityDialogProps) => {
+  const [completed, setCompleted] = useState<boolean | null>(null);
+  const [isUpdateSuccess, setIsUpdateSuccess] = useState<boolean>(false);
+  const [isMessageVisible, setIsMessageVisible] = useState<boolean>(false);
+  const [error, setError] = useState<boolean | null>(null);
+  const [isErrorShown, setIsErrorShown] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  useEffect(() => {
```
seidior, line 108: I'd love a line break between the variable declarations block and the `useEffect()` call to improve readability.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+    }
+    fetchData();
+  }, [contents, show]);
+
+  const sendCompletedStatus = (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+    completed: boolean
+  ) => {
+    event.preventDefault();
+    if (event.type !== 'click') return;
```
seidior, line 130: I'd swap lines 130 and 129 here, unless that breaks the functionality.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
-
     if (contents.start.toDateString() !== contents.end.toDateString()) {
       timeRangeString += formatDate(contents.end.toString()) + ' ';
     }
     timeRangeString += formatTime(contents.end.toString());
-
     return timeRangeString;
   };
-
+  const tableHeaderCellStyle = {
```
seidior, line 169: This can be moved outside our component and replaced here with a line break.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+        sx={{
+          textAlign: 'center',
+          fontWeight: 'bold',
+          display: 'flex',
+          alignItems: 'center',
+          justifyContent: 'center',
+        }}
       >
         {contents.title}
+        {completed && <TaskAltIcon sx={{ ml: 1 }} />}
```
seidior, line 194: I can't reproduce Prabh's screenshot because that activity no longer exists in the database, but if you wanted a workaround for his issue that doesn't require adding a background to the IconButton (which you shouldn't because a. that's not a good solution, and b. makes it harder for us to support Dark Mode if we ever decide to do that in the future), you could have the child of `<DialogTitle>` be a `<Stack>` and handle it that way:

```typescript
      <DialogTitle id="alert-dialog-title" sx={{ fontWeight: 'bold' }}>
        <Stack
          direction="row"
          justifyContent="space-between"
          alignItems="center"
          spacing={2}
        >
          <div></div>
          <div>
            {contents.title}
            {completed && <TaskAltIcon sx={{ ml: 1 }} />}
          </div>
          <IconButton
            aria-label="close"
            onClick={handleClose}
            sx={{
              color: theme => theme.palette.grey[500],
            }}
          >
            <CloseIcon />
          </IconButton>
        </Stack>
      </DialogTitle>
```

But! Up to you to decide which is the best approach here:

1. The advantage to the `position: absolute` approach is that the close button is always in the top right corner no matter what, but the disadvantages are that it can overlap anything that's behind it, and putting a background behind it may hide the activity completion icon.
2. The advantage to the <Stack> approach is that the close icon won't appear on top of anything, but the disadvantage is that I have no idea how it will behave with a long title.
3. Do nothing, ship as is.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -138,16 +272,69 @@ const ProgramActivityDialog = ({
...
+                    severity={isUpdateSuccess ? 'success' : 'error'}
+                    sx={{ width: '100%' }}
+                  >
+                    {isUpdateSuccess
+                      ? 'Assignment status updated successfully.'
+                      : 'There was an error updating the assignment status.'}
+                  </Alert>
+                </Snackbar>
+              )}
+              {isErrorShown && (
```
seidior, line 322: In general, <abbr>DRY</abbr> (don't repeat yourself) whenever possible when an overload of existing functionality could easily work, but I think in this case it's fine because we're going to remove lines 322 on when we consolidate loading the activity statuses all at once, so this is okay.

webapp/src/components/ProgramCalendar.tsx
```diff
@@ -53,7 +56,10 @@ const ProgramCalendar = ({
     allDay: false,
     start: new Date(),
```
seidior, line 57: I'm thinking the defaults for start and end maybe should have a different value, perhaps `new Date(0)`. I know it's not in scope for your ticket but while you're here anyway, might as well!

seidior (changes requested):
Tons of work put in by you and the rest of the team, and I really love how this has turned out. You looked beyond the scope of the issue and the exact wording to provide a great user experience, and the positive reaction from the rest of the team is evidence that it was worth the extra thought.

I have a few minor style points I'd like you to consider addressing, but they should be minor changes that shouldn't take too much time to address.

Thank you!

seidior:
@Prabhj33t I can't reproduce that screenshot that you have posted since your database is out of date; that activity doesn't exist anymore. Can you run the following commands and see if you can reproduce that behavior?

```shell
cd ~/Developer/rhizone-lms (or wherever you keep the repo)
git checkout 457-mark-task-completed-in-calendar
git pull
docker compose down && docker compose build --no-cache --pull && docker compose pull && docker compose up
```

Also,
> The x button overlaps with the check sign here otherwise, all seems good. I approve of this ticket

For pull request approvals, make sure to mark your comments as an "approval"; it defaults to leave the review as a "comment" status, which won't count for one of @wendyyng's two approvals needed, unfortunately.

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,212 @@
...
+const sendAPIPutRequest = (
+  path: string,
+  body: Object,
+  setCompleted: React.Dispatch<React.SetStateAction<boolean | null>>,
+  setIsUpdateSuccess: React.Dispatch<React.SetStateAction<boolean>>,
+  setIsMessageVisible: React.Dispatch<React.SetStateAction<boolean>>,
+  setIsLoading: React.Dispatch<React.SetStateAction<boolean>>
+) => {
+  setIsLoading(true);
+  if ('completed' in body && body.completed === null) {
```
phoebehala, line 71: wondering if only me got this warning 'Property 'completed' does not exist on type 'Object'.'

image

camianderson: It is showing on my machine too.

rrabinovitch: it's not appearing on my end

SanazDn:
Great work! All the tasks as expected.
_'Mark Complete/Incomplete' button works
_ mark activities as completed or incomplete by clicking the button works
_If I change the code to `fetch(`abc.com`` in ProgramActivityDialog in line 41 => it shows disable button with error "getting the assignment status"
_If change the code to "204" in programsRouter in line 102 => it shows error "updating the assignment status"

SanazDn (approved):

camianderson (changes requested):
I just have this two observations to be corrected.
Besides that you did a great job.
Everything is working as expected in my local enviroment.
Thank you.

rrabinovitch (approved):
the mark complete/incomplete button appears as expected in the activity dialog for activities of type: assignment, clicking the button toggles the checkmark icon next to the activity title, and the toast success/error notification works as expected.

thanks for taking the lead on this heavy ticket Wendy!

webapp/src/components/ProgramActivityDialog.tsx
```diff
@@ -1,85 +1,213 @@
...
+          setCompleted(data.completed);
+        }
+      },
+      error => {
+        setIsLoading(false);
+        setError(true);
+        setIsErrorShown(true);
+      }
+    );
+};
```
seidior, line 64: We've got a `.catch()` for the `fetch()` request in `sendAPIPutRequest`, so perhaps we should add one here too?

webapp/src/components/ProgramActivityDialog.tsx
wendyyng: Imports from the icons were grouped together and IconButton import is moved to the @mui/material

veranika-karpava:
Update 12/14:
- add space between defined variables and useEffect;
- add empty line above sendAPIGetRequest in ProgramActivityDialog;
- reorganize imports of icons from MUI in ProgramActivityDialog;
- import useState, useEffect, Dispatch, setStateAction from React in ProgramPage, ProgramActivityDialog and clean up code;
- specify the type of body of sendAPIPutRequest;
- remove async from sendAPIGetRequest in ProgramActivityDialog;
- import MouseEvent from React
- working on error handling

seidior (approved):
Squeezing some last minute changes in under the wire to help switch my review from "request changes" to "approve":

- I tried to make the GET/PUT functions in `ProgramActivityDialog.tsx` to look as close to one another as possible and have the same branching behavior, removing some duplicated function calls inside along the way. They're probably as close as they can get unless we want to copy the error message displaying functionality from @veranika-karpava to the PUT function too.
- I tackled the final three bits of unresolved feedback from comments I left earlier today, including switching out the default time values. This allowed me to pull the `timeRange()` function out of the component.
- I noticed some flickering after clicking the "Mark Completed" button that came from the fact that it switches to the loading state but then switches to "Mark Incomplete" very quickly. To fix that, I added a max 200ms delay to displaying the loading state after clicking the button. If the network request comes back within 200ms, it never switches the button to the loading state. If it takes longer than that, it's totally ok to show the loading status.

I don't... *love* that last bit of engineering, but it solves the problem. The more ideal way to handle this is with a custom MUI `<Transition>` element (in this case a custom `<Fade>`) that handles all the states of showing and hiding the Dialog. But I think this is good enough for now.

Test it out and let me know how it looks! If you like it, go ahead and merge the PR :)

camianderson (approved):
Great job! 
The changes are all working in my local machine as expected.
I am approving to be merged.

seidior:
Based on valuable feedback provided by @wendyyng this morning in standup, I've modified the error message display in the toast. In many cases, it will now be much shorter, looking a little something like this: 

shorter_error_toast

(Contrived example, obviously; the server wouldn't send a SyntaxError. But you get the point.)

Let me know if that adequately resolves the feedback, and if it does, let's merge this PR!

PR #487, opened 2022-12-13
469 api route and service function get all activities

### LINKED ISSUE #469: There should be an API route and service function that can get all activities for a program
## Details and resources

There should be an API route and service function to support getting all activities for a program and the participant's completed status for each of them, rather than having to get this information one-by-one for each relevant activity on the calendar.

This ticket is open for anyone who wants to take it.

This will include the route:

`[GET] /activityStatus/:programId`

And it will include a function in programsService.ts called something like `listParticipantActivitiesCompletionForProgram` that would take a participant ID and a program ID as a parameter and return a list of all assignments (activities of type "assignment") for the program with the participant's completed status, or it would return null if it can't find the program ID or the program has no associated assignments.

Here's what I had typed up previously, for reference:

```typescript
/**
 * Get the completion status (either true or false) of all activities of type assignment in a given program.
 *
 * @param {number} principalId - the unique id for the user
 * @param {number} programId - the id for the unique program
 * @returns List of activities and their completion statuses,
 *   or null if programId doesn't exist or it doesn't have activities
 *   of type "assignment"
 */
export const listParticipantActivitiesCompletionForProgram = async (
  principalId: number,
  programId: number
) => {
  // TODO: get list of all activities in program where activity_type_id is assignment (you'll need a join statement here)
  // TODO: get all `participant_activities` records for the principalId and programId
  // TODO: match up the two lists and generate the return list
  // TODO: return list should be an array of objects that look like this:
  // { programId: number, activityId: number, completed: boolean }
};
```

You will need appropriate tests for programsService.ts:

```typescript
describe('listParticipantActivitiesCompletionForProgram', () => {
  it('should return a list of participant activities with their completion statuses', async () => {});
  it('should return null if the programId does not exist', async () => {});
  it('should return null if the programId does not have any assignments', async () => {});
});
```

and for programsRouter.ts:

```typescript
describe('GET /activityStatus/:programId', () => {
  it('should return a list of activities and their completed statuses', done => {
    // TODO once this function is written.
  });
});
```

## Checklist

ChainShoe-Z:
## Proposed changes
This PR resolves #469 .

This PR adds a "listParticipantActivitiesCompletionForProgram" service function and a router for it. It takes the program and principle ID as parameters, and return all corresponding activities with their completion status.

Test files for both service function and router are 100% passed.

## Checklist

### DIFF
--- a/api/src/middleware/__tests__/programsRouter.ts
+++ b/api/src/middleware/__tests__/programsRouter.ts
@@ -3,6 +3,7 @@ import {
   listProgramsWithActivities,
   getParticipantActivityCompletion,
   setParticipantActivityCompletion,
+  listParticipantActivitiesCompletionForProgram,
 } from '../../services/programsService';
 import { createAppAgentForRouter, mockPrincipalId } from '../routerTestUtils';
 import programsRouter from '../programsRouter';
@@ -15,6 +16,9 @@ const mockGetParticipantActivityCompletion = jest.mocked(
 );
 const mockSetParticipantActivityCompletion = jest.mocked(
   setParticipantActivityCompletion
+);
+const mockListParticipantActivitiesCompletionForProgram = jest.mocked(
+  listParticipantActivitiesCompletionForProgram
 );
 
 describe('programsRouter', () => {
@@ -117,7 +121,6 @@ describe('programsRouter', () => {
       appAgent
         .get(`/activityStatus/${programId}/${activityId}`)
         .expect(400, done);
-      // should this include a check for the error message content? programsRouter L36
     });
 
     it('should respond with a bad request error if given an invalid activity id', done => {
@@ -127,7 +130,6 @@ describe('programsRouter', () => {
       appAgent
         .get(`/activityStatus/${programId}/${activityId}`)
         .expect(400, done);
-      // should this include a check for the error message content? programsRouter L36
     });
 
     it('should respond with an internal server error if an error was thrown while getting participant activity completion status', done => {
@@ -183,7 +185,6 @@ describe('programsRouter', () => {
           completed: true,
         })
         .expect(400, done);
-      // should this include a check for the error message content? programsRouter L79
     });
 
     it('should respond with a bad request error if given an invalid activity id', done => {
@@ -196,7 +197,6 @@ describe('programsRouter', () => {
           completed: true,
         })
         .expect(400, done);
-      // should this include a check for the error message content? programsRouter L79
     });
 
     it('should respond with an internal server error if an error was thrown while setting participant activity completion status', done => {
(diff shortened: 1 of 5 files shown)


api/src/services/programsService.ts
```diff
@@ -384,3 +384,24 @@ export const setParticipantActivityCompletion = async (
     completed: participantActivity.completed === 1,
   };
 };
+
+//  * Get the completion status (either true or false) of all activities of type assignment in a given program.
```
seidior, line 388: Your function header comments are not to JSDoc/TSDoc standard. Please adjust these to be conformant in the same way that the other function headers in this file are.

api/src/services/__tests__/programsService.ts
```diff
@@ -611,4 +623,26 @@ describe('programsService', () => {
       });
     });
   });
+
+  describe('listParticipantActivitiesCompletionForProgram', () => {
+    const programId = 1;
+    const principalId = 3;
```
seidior, line 629: These two `const` declarations are done outside of the test. I would recommend moving them inside of the test unless these values are shared by multiple tests.

api/src/services/__tests__/programsService.ts
```diff
@@ -611,4 +623,26 @@ describe('programsService', () => {
...
+    it('should return a list of participant activities with their completion statuses', async () => {
+      mockQuery(
+        'select `activity_id`, `completed` from `participant_activities` where `program_id` = ? and `principal_id` = ?',
+        [programId, principalId],
+        [
+          { activity_id: 7, completed: false },
```
seidior, line 635: Perhaps there's a way to extract this information from `participantActivitiesForProgram` (declared on line 174) instead of repeating their declaration here?

api/src/services/__tests__/programsService.ts
```diff
@@ -168,6 +170,16 @@ const participantActivitiesList: ParticipantActivities[] = [
     completed: true,
   },
 ];
+
+const participantActivitiesForProgram: participantAcitivityForProgram = {
+  programId: 1,
+  participantActivities: [
+    { activity_id: 7, completed: false },
```
seidior, line 177: Perhaps these values can be extracted from `participantActivitiesList` instead of re-declaring them?

api/src/models.d.ts
```diff
@@ -62,3 +62,13 @@ export interface ParticipantActivities {
   principal_id: number;
   completed: boolean;
 }
+
+export interface ParticipantActivitiesWithCompletionStatus {
+  activity_id: number;
+  completed: boolean;
+}
+
+export interface participantAcitivityForProgram {
```
seidior, line 71: This interface is misspelled, and it should be capitalized.

api/src/models.d.ts
```diff
@@ -62,3 +62,13 @@ export interface ParticipantActivities {
   principal_id: number;
   completed: boolean;
 }
+
+export interface ParticipantActivitiesWithCompletionStatus {
```
seidior, line 66: This object structure defines a single activity and its completed status, so I think a better name for this would be `ParticipantActivityCompletionStatus`, switching to the singular form of activity.

Then, when you have multiple objects in an array it would read as `ParticipantActivityCompletionStatus[]`, which makes more sense since it's an array of singular objects.

api/src/models.d.ts
```diff
@@ -62,3 +62,13 @@ export interface ParticipantActivities {
...
+  activity_id: number;
+  completed: boolean;
+}
+
+export interface participantAcitivityForProgram {
+  programId: number;
```
seidior, line 72: Every other object member defined in this file uses the underscore version (`created_at`, `updated_at`, etc) and not the camelCase version like you have here. I would suggest changing these to `program_id` and `participant_activities`.

api/src/middleware/__tests__/programsRouter.ts
```diff
@@ -196,7 +197,6 @@ describe('programsRouter', () => {
           completed: true,
         })
         .expect(400, done);
-      // should this include a check for the error message content? programsRouter L79
```
seidior, line 199: Thank you for removing these comments!

api/src/middleware/__tests__/programsRouter.ts
```diff
@@ -211,4 +211,48 @@ describe('programsRouter', () => {
         .expect(500, done);
     });
   });
+
+  describe('GET /activityStatus/:programId', () => {
```
seidior, line 215: The tests should appear in the same order in the test file that the routes do in the router middleware file; in the middleware file, your route is defined *above* the route that specifies both the `programId` and the `activityId`; you should move your tests (defined starting on line 215 of this file) *above* the tests for the route that specifies both the `programId` and `activityId`.

api/src/middleware/__tests__/programsRouter.ts
```diff
@@ -211,4 +211,48 @@ describe('programsRouter', () => {
...
+    it('should return a list of activities and their completed statuses', done => {
```
seidior, line 216: It should return the program ID and a list of the program's activities and the principal's completed statuses.

seidior (changes requested):
Overall, great work in getting this done and I think you're almost ready to go here. However, I'm requesting you make some changes before it meets merger approval.

That said, I rebuilt the Docker containers without issue, and it passes my `yarn delint` and `yarn test` tests.

phoebehala (approved):
Nice work! I tested the below and it works fine on my end!
-  `cd api` and `yarn test`: all test passed
-  `cd api` and `yarn delint`: no files changed
-   Tested the get rout and it returned {"data":{"program_id":1,"participant_activities":[]}} (Noted there is no default rows in participant_activities table)

SanazDn (approved):
Great job! All tests passed with yarn test.

seidior (approved):
With the last few changes, I can now approve this pull request. Great work everyone!

ChainShoe-Z:
Merge!

PR #493, opened 2022-12-15
#470 Get all assignments and their completion at once in webapp

### LINKED ISSUE #470: The frontend should be adapted to use an API call that gets all assignments and their completion
## Details and resources

As per issue #469, there will be an API route that will allow us to get the completed status of all assignments in a program at once. Right now in the webapp, we are retrieving this information one-by-one as needed.

In this issue, we will compile all work required to adapt the webapp to this new route for fetching the information. (The existing logic to set a new value for a completed status for an assignment should be otherwise unchanged, although if any changes to where we are storing assignments' completed statuses is required, then the setting logic may have to be updated for the state to persist.)

This issue is open for anyone who wants to tackle it.

## Checklist

phoebehala:
## Proposed changes

This resolves #470 by:
- Retrieving the completed status of all assignments in a program at once.
- Addressing a type issue with the return type from a backend service file function.

## Checklist

### DIFF
--- a/webapp/src/components/ProgramCalendar.tsx
+++ b/webapp/src/components/ProgramCalendar.tsx
@@ -51,30 +51,40 @@ const ProgramCalendar = ({
   setCurrentView,
   viewOptions,
 }: ProgramCalendarProps) => {
-console.log(program.activities); 
-const {
-  data: participantActivitiesCompletionList,
-  error,
-  isLoading,
-} = useApiData<participantActivitiesCompletionList>({
-  deps: [],
-  path: `/programs/activityStatus/${selectedProgram+1}`,
-  sendCredentials: true,
-});
-console.log( participantActivitiesCompletionList);
-console.log(program.activities);
-if(participantActivitiesCompletionList){
-  for(var i = 0; i < program.activities.length; i++){
-    for(var j = 0; j < participantActivitiesCompletionList?.participantActivities.length; j++){ 
-        if(program.activities[i].curriculum_activity_id=== participantActivitiesCompletionList.participantActivities[j].activity_id){
-          program.activities[i].completed = participantActivitiesCompletionList.participantActivities[j].completed === 1
-        }else{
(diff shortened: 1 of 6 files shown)


webapp/src/components/ProgramCalendar.tsx
```diff
@@ -65,6 +75,54 @@ const ProgramCalendar = ({
...
+        } else {
+          return calendarActivity;
+        }
+      }
+    );
+    setCalendarActivities(updatedCalendarActivities);
+  };
+
+  const { data: participantActivitiesCompletionList, error } =
+    useApiData<ParticipantActivityForProgram>({
```
phoebehala, line 101: can we do a little tweak here to make it like other API call
`  const { 
    data: participantActivitiesCompletionList, 
    error 
  } = useApiData<ParticipantActivityForProgram>({`

webapp/src/components/ProgramCalendar.tsx
```diff
@@ -65,6 +75,54 @@ const ProgramCalendar = ({
   const [dialogContents, setDialogContents] = useState<CalendarEvent>(
     defaultDialogContents
   );
+  const [activitiesCompletion, setActivitiesCompletion] =
+    React.useState<ParticipantActivityForProgram>({
```
veranika-karpava, line 79: One more React I didn't delete. Please delete it

webapp/src/components/ProgramCalendar.tsx
seidior: Unfortunately this will cause our `yarn delint` checks to fail :( good catch though!

webapp/src/components/ProgramCalendar.tsx
seidior: Good catch, thanks!

veranika-karpava (approved):
Everything works as expected on my machine. 
All test and delint passed.
Amazing work, @phoebehala!!!

PR #547, opened 2023-03-15
#536 Add tests for get assessments route

### LINKED ISSUE #536: There needs to be a test for the GET /assessments route
### Describe the Feature / Enhancement

There needs to be a test for the GET `/assessments` route (from the router file), which currently does not exist. We need to test this route to ensure the code written for the test will behave as expected by the protocol we previously defined.

This issue continues the work from issue #515.

### Additional Details and Resources

This route has no URL parameters and expects no body in the request. The response should contain the following:

#### Facilitator:
```typescript
collectionEnvelope([{
  curriculum_assessment: CurriculumAssessment;
  program_assessment: ProgramAssessment;
  submissions_summary: FacilitatorAssessmentSubmissionsSummary;
}])
```

#### Participant:
```typescript
collectionEnvelope([{
  curriculum_assessment: CurriculumAssessment;
  program_assessment: ProgramAssessment;
  submissions_summary: AssessmentSubmissionsSummary;
}])
```

### Correct Project Selected

- [X] This issue is linked to the appropriate project

### Labels

- [X] This issue has appropriate labels.

Awatanka:
## Proposed changes

This pull request resolves #536 by adding tests for the `GET /assessments` route (from the router file). 

## Checklist

- [x] Related issue appears at beginning of pull request title with pound sign, and title describes the changes being proposed.
- [x] Related issue is linked in pull request description using [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).
- [x] The appropriate label has been chosen for this pull request.
- [x] The correct project has been selected for this pull request.
- [x] All commits in this branch, including merge commits, begin with the issue number and a pound sign.
- [x] Tests have been added, where appropriate; or, no tests are relevant for this pull request.
- [x] This pull request contains UI changes, and screenshots of those UI images appear below; or, this pull request contains no UI changes.


api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,263 @@
...
+            done(err);
+          }
+        );
+    });
+
+    it('should respond with a list of all assessments (without questions) for participant enrolled in one program', done => {
+      mockPrincipalId(enrolledParticipantPrincipalId);
+      const participantAssessmentListResponse = [
+        {
+          curriculum_assessment:
```
seidior: Every test in a test file is run asynchronously, so when you have code that is executing in a test that can affect other tests (as you do here), when we go to run this code, you will have unintended side effects.

Is there any particular reason you've written it like this, or can we write it like:
```typescript
const participantAssessmentListResponse = [{
  curriculum_assessment: curriculumAssessment,
  program_assessment: programAssessment,
  submissions_summary: assessmentSubmissionsSummary
}];
```

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,263 @@
...
+            // call a function that returns the permission of the user for each program (participant/facilitator)
+            expect(mockGetProgramIdByProgramAssessmentId).toHaveBeenCalledWith(
+              programAssessmentId
+            );
+            expect(mockFindRoleInProgram).toHaveBeenCalledWith(
+              programId,
+              enrolledParticipantPrincipalId
+            );
+            // get a list of program assessments for each program the user is enrolled in
+            expect(mockPrincipalEnrolledPrograms).toHaveBeenCalledWith(
```
seidior: Why would we be calling this function twice?

api/src/models.d.ts
```diff
@@ -159,14 +159,14 @@ export interface AssessmentSubmission {
...
 }
 
-export interface SubmittedAssessment {
+export interface AssessmentWithSubmissions {
   curriculum_assessment: CurriculumAssessment;
   program_assessment: ProgramAssessment;
   submissions: AssessmentSubmission[];
 }
 
-export interface ProgramParticipantCompletionSummary {
```
seidior: How come we're removing this model?

seidior (changes requested):
All in all, great work @Awatanka! You're very nearly there. I just have a few comments/suggestions to address first.

Also, remember to run `yarn delint` before committing and pushing to GitHub. I know we won't have automated checks on these testing PRs to highlight it so it will involve manually remembering this, but it's something I'd like to see regardless :)

api/src/middleware/__tests__/assessmentsRouter.ts
Awatanka: Yes, I see your point. I will add changes.

api/src/models.d.ts
Awatanka: I have fixed it. Thanks

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+const curriculumAssessment: CurriculumAssessment = {
+  id: curriculumAssessmentId,
+  title: 'Assignment 1: React',
+  description: 'Your assignment for week 1 learning.',
+  max_score: 10,
+  max_num_submissions: 3,
+  time_limit: 120,
+  curriculum_id: 3,
+  activity_id: 97,
+  principal_id: facilitatorId || participantId,
```
seidior: I don't think you mean to be using a logical OR here.

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+            participantAssessmentListResponse.length
+          ),
+          err => {
+            // call a (mock) function that gets a list of programs the user is enrolled in
+            expect(mockPrincipalEnrolledPrograms).toHaveBeenCalledWith(
+              enrolledParticipantPrincipalId
+            );
+            // call a function that returns the permission of the user for each program (participant/facilitator)
+            expect(mockGetProgramIdByProgramAssessmentId).toHaveBeenCalledWith(
+              programAssessmentId
```
seidior: The only information you would have access to by the time the flow of the test gets to line 148 would be the following:

- `principalId` of the user making the request
- the return value from your first function (`principalEnrolledPrograms`), which you indicate from line 121 is `[4]`

So, how can we call `getProgramIdByProgramAssessmentId` if neither of these things we have access to include the `programAssessmentId`?

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+      ];
+      // mock response from (function that gets a list of programs the user is enrolled in) to include one program
+      mockPrincipalEnrolledPrograms.mockResolvedValue([4]);
+      // mock response from (call a function that returns the permission of the user for each program (participant/facilitator)) to respond with participant for that one program
+      mockGetProgramIdByProgramAssessmentId.mockResolvedValue([
+        exampleAssessmentId,
+      ]);
+      mockFindRoleInProgram.mockResolvedValue({ title: 'participant' });
+      // mock response from (get a list of program assessments for each program the user is enrolled in) to respond with a list of program assessments
+      mockGetAssessmentsForProgram.mockResolvedValue(programAssessment);
```
seidior: I would imagine this function would return an array of objects rather than a singular object, no?

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+              programId,
+              enrolledParticipantPrincipalId
+            );
+            //get a list of program assessments for each program the user is enrolled in) to respond with a list of program assessments
+            expect(mockGetAssessmentsForProgram).toHaveBeenCalledWith(
+              programId
+            );
+            // get a list of curriculum assessments that correspond to each program assessment
+            expect(mockGetCurriculumAssessmentById).toHaveBeenCalledWith(
+              exampleAssessmentId,
```
seidior: This first parameter should be `programAssessment.assessment_id` instead. Even if it's the same value, it makes more sense here, since we should be getting back an array that includes `programAssessment` from the `getAssessmentsForProgram` function.

seidior (changes requested):
A couple more items for your review! The comments apply to both tests, so when you resolve the issues with the first one, apply the same to the second one.

Let me know if you have questions!

api/src/middleware/__tests__/assessmentsRouter.ts
Awatanka: Yes, I missed that. Thanks

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+      mockGetProgramIdByProgramAssessmentId.mockResolvedValue([2]);
+      mockFindRoleInProgram.mockResolvedValue({ title: 'facilitator' });
+      // mock response from (get a list of program assessments for each program the user is enrolled in) to respond with a list of program assessments
+      mockGetAssessmentsForProgram.mockResolvedValue(programAssessment);
+      // mock responses from (get a list of curriculum assessments that correspond to each program assessment) to respond with the corresponding curriculum assessment for each program assessment in that previous list
+      mockGetCurriculumAssessmentById.mockResolvedValue(curriculumAssessment);
+      // mock responses from (call a (mock) function that gets the facilitator assessment summary for each program assessment where the user is a facilitator of that program) to respond with the assessment summary for all participants for each program assessment
+      mockGetFacilitatorAssessmentSubmissionsSummary.mockResolvedValue(
+        facilitatorAssessmentSubmissionsSummary
```
mcharrod, line 209: We'll want to take out the commented code before merging I think

api/src/middleware/__tests__/assessmentsRouter.ts
```diff
@@ -1,18 +1,242 @@
...
+    const enrolledParticipantPrincipalId = 5;
+    const facilitatorPrincipalId = 6;
+    const programIds: number[] = [];
+
+    it('should respond with an empty list for a user not enrolled in any programs', done => {
+      // mock response from function that gets a list of programs the user is enrolled in to return an empty list
+      mockPrincipalEnrolledPrograms.mockResolvedValue(programIds);
+
+      appAgent
+        .get('/')
```
mcharrod, line 118: It would help if you could be more specific on the list of what we are getting here. Are we trying to get a list of programs the mentee is enrolled in? Also, I think that the get '/' path will need to be updated for the path we are trying to go to.

mcharrod (commented):
Thanks for writing these tests! They are easy to read and understand. I do think the paths need to be updated to reflect where we're trying to go, unless I'm misunderstanding something in the code. Thanks again!

api/src/middleware/__tests__/assessmentsRouter.ts
Awatanka: Thanks, I will.

api/src/middleware/__tests__/assessmentsRouter.ts
Awatanka: Hi, quick question. To be able to use an array here like this `mockGetAssessmentsForProgram.mockResolvedValue([programAssessment])`, I need to change the promise in the `getAssessmentsForProgram` function.  It returns just an object. So, I am not sure that I am able to resolve this suggestion without changing `assessmentService.ts` file.

seidior: Understood; we'll definitely need to change that function in the next ticket. For now, I'm approving you to change this to `mockGetAssessmentsForProgram.mockResolvedValue([programAssessment])`, even if it gives you an error in VSCode.

seidior (approved):
Thank you for addressing my feedback. You're welcome to merge after you bring your branch up to date!

